PR #19, opened 2019-12-11
Add ScrollRegion

dhardy:
This requires a few changes:

- support multiple draw passes
- support for clipping drawables to a sub-region (via a new draw pass to let us offload the work to the GPU; inspired by [Iced](https://github.com/hecrj/iced/))
- support for offsetting drawables
- the new widget
- (slightly hacky) support for handling scroll events

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ version = "0.8"
 
 [dependencies.winit]
 # Provides translations for several winit types
-version = "0.20.0-alpha3"
+version = "0.20.0-alpha5"
 optional = true
 
 [workspace]
--- a/kas-macros/src/layout.rs
+++ b/kas-macros/src/layout.rs
@@ -95,13 +95,14 @@ pub(crate) fn derive(
             fn draw(
                 &self,
                 draw_handle: &mut dyn kas::theme::DrawHandle,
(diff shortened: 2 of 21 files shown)


dhardy:
Issue: fonts are not clipped to scroll windows. This is because `glyph_brush` draws everything in a single render pass, which doesn't let us apply GPU scissor regions as required. Options:

- Queue sections for each region independently, using a new render pass after each. This is what Iced does, but is not quite so easy with KAS's render pipe. This also implies one should use `GlyphBrush::keep_cached` to keep sections in the cache (which Iced currently doesn't; hint @hecrj).
- Use a separate instance of `GlyphBrush` for each region. This makes it easy to queue sections correctly, but requires constructing new instances on demand (may cause lag) and prevents sharing of cached glyphs.
- Use the `z` value included in sections and assign a distinct depth value to each layer. Needs further investigation.
- Perform clipping within `wgpu_glyph` by adding a bounding rect to each section.

This turned out more complex than I had expected. @hecrj may I ask if you put any effort into investigating alternatives here?

dhardy:
Decision: merge with noted caveats (the issues just opened). Most of this PR is useful and I don't wish to block development (or push everything into this PR).

PR #24, opened 2019-12-16
Add Dynamic widget parent

dhardy:
Adds `DynList` widget storing a `Vec<Box<dyn Widget>>`.

Some adjustments to `kas::layout` module for row/column solvers.


dhardy:
Also adjusted grid solvers with dynamic storage option. (This doesn't include a "dynamic grid" parent widget, but it should be possible to build one over the included APIs.)

dhardy:
This also implements `Widget` for `Box<dyn Widget>`.

Caveat: the `DynVec` and `DynList` widgets have a very limited API for interacting with child widgets (`push` and `resize_with`). This will need to be expanded in the future.

PR #30, opened 2019-12-17
Hacks for usage with stable Rust

### LINKED ISSUE #25: Target: stable Rust
KAS currently requires unstable language features, thus requires nightly Rust.

Features currently in use:

-   `new_uninit` — used to implement `Clone` for `Box<dyn Widget>`. I *think* we could do the same thing (in a "more unsafe" way) using transmutes, but haven't tested.
-   `proc_macro_hygiene` — used in all examples to use the `make_widget!` macro; we should be able to work around this using [proc-macro-hack](https://github.com/dtolnay/proc-macro-hack)

Features *not* in use but highly desired:

-   const generics: this would massively simplify the layout storage types and usage
-   specialisation: this would allow nicer code around `Response` conversions

*Both* the above features have partial nightly implementations, and in *both* cases these implementations have caused an ICE when attempting to use them to address the above. (Perhaps this is for the best for now, since stable-Rust workarounds are possible.)

dhardy:
This puts proc-macro diagnostics behind a feature gate, and uses [proc-macro-hack](https://github.com/dtolnay/proc-macro-hack) to avoid a requirement on `proc_macro_hygiene`.

Fixes #25.

I am **unsure whether to merge this**. As nice as it is to be able to use stable Rust, it hides macro diagnostics by default, uglifies the code-base (especially with `make_widget` having to be exported from the root of `kas`), and examples need high macro recursion limits:
```rust
// calculator.rs
#![recursion_limit = "1024"]
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,10 @@ version = "0.8"
 version = "0.20.0-alpha5"
 optional = true
 
+[dependencies]
+proc-macro-hack = "0.5"
+proc-macro-nested = "0.1"
+
 [workspace]
 members = ["kas-macros", "kas-wgpu"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ repository = "https://github.com/dhardy/kas"
 
 [features]
 # Enables usage of unstable Rust features
-nightly = []
+nightly = ["kas-macros/nightly"]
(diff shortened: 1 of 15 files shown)


dhardy:
This is now completely obsolete (thanks Rust 1.45 :tada:).

Issue #28, opened 2019-12-17
Dynamic window creation, modification, deletion


dhardy:
Currently, all windows must be added before the event loop is started and cannot be modified. It should be possible to:

- [x] add windows from within event-handlers
- [x] inspect window state between window-close and state-deletion
- [x] ~modify running windows (e.g. to adjust font properties)~

dhardy:
> modify running windows (e.g. to adjust font properties)

Crossing this one off the list: it can either be seen as run-time theme adjustment or as per-window theme support, which should be a different issue.

PR #36, opened 2020-01-16
Add ToolkitProxy, WindowId, closure callbacks

dhardy:
First, this adds `ToolkitProxy`, allowing windows to be closed from other threads.

Second, this adds a `WindowId` type, allowing specific windows to be identified.

Third, this implements the final step in #28: closure callbacks. (It does so via two different closure types, which is somewhat redundant.)


dhardy:
Fourth, this makes `TkWindow` a temporary wrapper with a borrow of `SharedState` and adds a `pending` list to the latter.

Fifth, this moves window configuration to after init, avoiding the need for reconfiguration where init callbacks adjust things.

Overall, it significantly revises `Loop::handle`, and fixes some subtle bugs where requested actions were ignored.

PR #47, opened 2020-01-30
Fix region moved under the mouse

dhardy:


dhardy:
This PR fixes a subtle bug: when using the mouse wheel over a scrollable region, the mouse-hover target was not previously updated until the mouse moved (affecting highlighting and clicks on widgets).

Also, this PR finally adds a way to find a child widget by coordinate without requiring a `Manager` / event handler responses.

PR #54, opened 2020-02-13
Cursor icons

### LINKED ISSUE #45: Set cursor icon
Event handlers should be able to set the mouse cursor icon. Probably needs an addition to `TkWindow`.

API: https://docs.rs/winit/0.20.0/winit/window/struct.Window.html#method.set_cursor_icon

dhardy:
Small PR to enable setting mouse cursor icon during hover and grab. Closes #45.

Currently this simply pulls in the `winit::window::CursorIcon` enum. Eventually we may want something else, but for now this will do.


dhardy:
Worth noting that I tried setting custom icons for hovering over scrollbars, but ultimately decided against it. At least on Linux, this seems to be the standard behaviour.

Even more interesting, we *could* change the icon when over any part of the UI that supports dragging inside a scroll region, however this would probably also be more annoying than useful (in any case, the icon changes when a grab is active).

PR #74, opened 2020-03-19
Trait revision (Layout & Handler), CowString, update actions

dhardy:
Adds `CowString` and `CowStringL` aliases: string handling now uses these.

`RadioBox`, `CheckBox` and `EditBox` have had their generics modified and `Handler` implementations unified.

Adjust traits:

- some properties moved to `WidgetCore` with derive support
- rename `Handler::handle_action → action`
- move `Handler::handle → Layout::event`
- make `Layout: Handler`
- macro adjustments associated with these changes
- replace `update_timer` and `update_handle` methods with `Action` values (this allows the handlers to return a message to the parent widget)

Much of the above is motivated by the `ComboBox` implementation, which is in progress.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,7 @@ script:
   - cargo test --all-features
   - cargo test --manifest-path kas-macros/Cargo.toml --all-features
   - cargo test --manifest-path kas-theme/Cargo.toml --all-features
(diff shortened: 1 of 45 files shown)


dhardy:
Both OS X runs timed out. If this also happens with the next PR, I'll just disable the OS X tests.

PR #84, opened 2020-04-10
Implement disabled and error states for widgets

### LINKED ISSUE #82: disabled and error states
- [x] make it possible to disable any input widget
- [x] distinct graphics for disabled input widgets

Since this affects many widgets, possibly it should be part of `WidgetCore` / `CoreData`. Effects:

- widgets should not be highlightable
- keyboard navigation should skip the widget
- active pointer/character grabs should be cancelled
- `Handler::action` should not be called?

Additionally, at least char-input fields should support an error state, causing the background to be highlighted red:

- [x] allow `EditBox` to be able to be put into an error state

dhardy:
Closes #82.

Omission: keyboard navigation still iterates through disabled widgets. Since the disabled state isn't really inherited (except to block event passing and in draw()), the obvious change to `Manager::next_nav_focus` doesn't work. TODO.


dhardy:
Omission fixed (via rebase): `fn find` and `fn find_id` now abort if disabled, instead of `EventHandler::event`.

PR #94, opened 2020-04-27
Depth buffer, WGPU update, misc

dhardy:
This adds a depth buffer, which is only written to by most draw pipes but used for depth-testing when rendering text (and optionally with custom pipes). This lets us prevent text from spilling over pop-ups and scroll regions. Ultimately we could stick with this approach or could use indirect rendering for pop-ups (and possibly also scroll regions).

Also:

- update to wgpu 0.5, font-kit 0.6 and smallvec 1.4
- fix visible hole inside flat-theme menu frame
- improve nav focus for ComboBox (initial and under cursor)
- fix pop-up position for a pop-up spawned from a widget inside a ScrollRegion

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ internal_doc = []
 conv = "0.3"
 log = "0.4"
 rusttype = "0.8"
-smallvec = "1.1"
+smallvec = "1.4"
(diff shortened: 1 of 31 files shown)


dhardy:
Not obvious what the error on macos is — I'll ignore for now.

PR #113, opened 2020-07-17
Shaping + Text-navigation; Rust 1.45 update

dhardy:
KAS-text update: implement most text navigation (Home, Ctrl+Home, Ctrl+Left, Up, PageDown, etc.)
The Ctrl+Left/Right part isn't quite right, but left/right navigation needs re-writing when we get to bidirectional text support anyway.

Additionally: with the Rust 1.45 release we no longer require proc_macro_hygiene!

Depends on: https://github.com/kas-gui/kas-text/pull/6

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ path = "kas-macros"
 
 [dependencies.kas-text]
 git = "https://github.com/kas-gui/kas-text"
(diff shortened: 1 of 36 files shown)


dhardy:
Oops, just made this PR larger than intended. No matter.

PR #115, opened 2020-08-05
Update kas-text (navigation)

dhardy:
Multi-line text editing now works fairly well, even with bidirectional content!

There are still several major limitations:

-   the design doesn't scale to large documents (noticeable when editing a couple hundred lines)
-   there is no implicit support for scrolling when texts don't fit in the edit box
-   short-cuts aren't handled aside from a few already mapped to control-codes by the OS (so e.g. Ctrl+V works on KDE/Linux but not Ctrl+A)
-   the undo buffer is limited to a single action and handled directly by `EditBox`
-   text formatting / highlighting (other than background) isn't started yet

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ path = "kas-macros"
 
 [dependencies.kas-text]
 git = "https://github.com/kas-gui/kas-text"
-rev = "b9e19c2105d9734ba396841d4d0e64ae688584b0"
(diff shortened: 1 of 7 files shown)


dhardy:
The short-cut limitation has been resolved, though without configuration yet. (Direct push to master, so no PR.)

dhardy:
Scrolling was enabled by #116; no scroll-bars yet (and possibly that should be a different widget).

PR #124, opened 2020-09-27
Use pre-compiled shaders

### LINKED ISSUE #123: Make shaderc dependency optional
Hello!

We discussed a bit in [the 3D widget of Iced](https://github.com/hecrj/iced/issues/343) I started to experiment with KAS. So far I struggled a bit to compile it and run it because of the ``shaderc`` dependencies. As the Rust crate is a wrapper around a C++ lib, some C++ tools make it somewhat difficult to built. So far I run into these case:
* On Windows, when I try to build the KAS, ``shaderc`` wants ninja to build it. As I don't have it the build fail.
* On Linux, I tried to cross compile it to Windows using mingw. ``shaderc`` use C++ 11 and for some reason the mingw tools I have on Linux doesn't seem to support it. I did not dive into it so it may be a incorrect use of me.

Anyway, I find that the use of ``shaderc`` make the compilation harder. I tried to compile KAS without ``shaderc`` and it was very easy. Basically, what I did was compile the existing shaders to SPIR-V using this [online tool](https://alexaltea.github.io/glslang.js). I also edited the ShaderManager from 
````rust
let vert_3122 = compile!(device, Vertex, "shaders_bin/scaled3122.vert");
````

to

````rust
let vert_3122 = device.create_shader_module(wgpu::include_spirv!("shaders_spir-v/scaled3122.vert"))
````
and it just worked.

So, my point is that since ``shaderc`` make the compilation less reliable it could be made an optional feature. That way, people who are OK with compiling their shaders before compile time can do it easily don't have to deal with ``shaderc``. The other that want a shader compiler that can be used at runtime can enable the flag and use it as today.

What do you think?

### PR DESCRIPTION
**Pull request by dhardy**
Fixes #123 

Instead of using the `shaderc` dependency (hard to install; #123 is not the first report), we pre-compile shaders with `glslc`. This is still part of the `shaderc` project, but can be installed from packages on at least Fedora and is only required when shaders are out-of-date.

Tested on Fedora; builds missing and out-of-date shaders only. Not tested on other platforms.

@hecrj can I get your thoughts — do you just compile by hand?

### CODE REVIEW

**Comment by dhardy** [comment]
@codec-abc could you test this please?

**Comment by dhardy** [comment]
Automatic compilation is now opt-in. If not enabled and re-compilation is required, the following warning message is shown and the build continues (using out-dated SPIR-V files):
```
warning: Shader compilation required: /home/dhardy/projects/gui/kas/kas-wgpu/src/draw/shaders/scaled32.vert
warning: No shader found. If you have a shader compiler such as glslc installed, try setting SHADERC=glslc
```

codec-abc:
For the record, this build and run just fine on my side. I Just say it again here because I replied in the issue #123 instead.

hecrj:
> @hecrj can I get your thoughts — do you just compile by hand?

Yes, I compile the SPIR-V files manually. I do not change shaders very often, so I haven't spent any effort in automating it yet.

PR #130, opened 2020-10-08
Replace Window callbacks with on_drop method returning a future

dhardy:
Use a simple future to allow safely recovering data from a child window on closure.

We don't use `std::future::Future` since its scheduling mechanism (waker and executor) is incompatible with our requirements, and anyway, this is simpler.

@codec-abc can I get you to do a code-review, at least of the changes to the gallery app? `TextEditPopup` (or a window wrapper) could be added to the standard dialog set, but possibly its too simple to bother with.

text-edit-popup


codec-abc:
Hey!

This is pretty great! Thanks for doing this. I really appreciate the effort.

I am not very good at code review but here are the only nitpicking points I found (as otherwise this look very fine to me)

1. The term Future is a little confusing. I think most of people who will think of the Future of the std. Naming is hard but it could be better, in my opinion, to find a name that convey the same idea but without using Future so it will be obvious it is not tied to the Rust async ecosystem.

2. Regarding the integration the TextEditPopup to the standard dialog set, it might indeed be too simple. On the other hand, I wonder if it possible to make a Ok/Close generic popup in the standard dialog set. It is not necessary that useful since most of the "boilerplate" code is in the code that create the popup rather than in the popup's code itself. But this way we will get a kind of standard behavior and look classic popups.

dhardy:
Yes, I'm not sold on the name `SimpleFuture`, but I'm uninspired regarding better names. But essentially it is a "future": a promise of a value which will eventually arrive. Just not part of the std async ecosystem.

dhardy:
I renamed `SimpleFuture` → `Future`. Perhaps this is even worse from the point-of-view of confusion with those in libstd, but [futures are a much older concept](https://en.wikipedia.org/wiki/Futures_and_promises). The name seems apt, but if someone has a better one I'll listen.

codec-abc:
I cannot think of another name too. In the end, it doesn't matter that much because the documentation makes it clear that the KAS futures are not related to Rust std's futures. Plus, you added an example of it so if it should clear the confusion for the one that skipped the documentation.

Issue #136, opened 2020-11-04
Pointer/touch behaviour


dhardy:
In KAS, pointer/touch event behaviour is handled by widgets:

- the obvious click-to-activate on buttons, checkboxes, etc.
- the usual click-drag or hover to navigate menus
- click-drag and touch-drag on "pan" widgets, e.g. Mandlebrot explorer
- click-drag and touch-drag on empty space inside `ScrollRegion` to pan
- touch-drag and ctrl+click-drag to pan text-edits
- click-drag and touch-wait-drag in text-edits to select text

Most of the above are mostly standard UI behaviours with a couple of exceptions (ctrl+click-drag to pan in text-edits and click-drag to pan in empty space). Questions:

1.  Do we want to keep the non-standard ctrl+click-drag to pan? Ideally yes, but possibly with a configuration option.
2.  Alternatively we could emulate touch behaviour: click-wait-drag to select text. But definitely not as default behaviour.
3.  If we allow text-selection within labels, do we allow the same behaviour as above? The problem is that `ScrollRegion` currently supports panning by touch/click drag on any widget in the region which does not handle this event; if `Label` starts handling clicks for text selection, this behaviour will be mostly unavailable. We could make this compatible with the Ctrl+click behaviour to support both, but possibly not with the timer behaviour (currently the widget has to choose whether to handle a click or not immediately).

dhardy:
This is all resolved:

1. Behaviour remains like this by default, but is configurable.
2. No: mouse does not use click+wait+drag for selection and I don't think this is wanted.
3. A special widget, `ScrollLabel`, allows text selection. Ctrl+drag allows pan behaviour (configurable).

PR #138, opened 2020-11-04
Label text selection

dhardy:
Extends #137 allowing text selection in labels, where the parent doesn't steal events (i.e. not in `CheckBox` and `RadioBox`). This impl works with caveats:

- mouse-selection is implemented but not touch selection (see #136)
- there is no keyboard control since Labels cannot have nav focus or char focus (possible but pretty sure we don't want this)
- `kas-text`'s `TextDisplay::text_index_nearest` and `highlight_lines` methods are a bit buggy on formatted text

Thus it is not possible to copy the selected text to the clipboard. A context menu or global helper for shortcuts like Ctrl+C should fix this.

Currently undecided whether to merge this (with fixes) or abandon it (at least for now).

### DIFF
--- a/kas-theme/src/dim.rs
+++ b/kas-theme/src/dim.rs
@@ -194,8 +194,11 @@ impl<'a> draw::SizeHandle for SizeHandle<'a> {
             SizeRules::new(min, ideal, margins, policy)
         } else {
             let min = match class {
+                TextClass::Label => required.1 as u32,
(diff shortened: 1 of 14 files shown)


dhardy:
#192 added a separate `ScrollLabel` widget which supports selection. It has one of the same limitations (no keyboard control), but is overall probably better: many labels such as those used on check/radio boxes and simple field labels would not be expected to support selection anyway (with the advantage that click+drag then works).

PR #164, opened 2021-02-23
View widgets: trait revision, reliable keys

dhardy:
This replaces the `Accessor` trait with `SingleData`, `SingleDataMut`, `ListData`, `ListDataMut`. In a way these are simpler; they also have a type of iterator (a ridiculous vec generator until GATs allow real iterators).

Keys are now the same for a list and its filtered version, so list selection now works properly on filtered lists.

Also moves `derive(Widget)` before its attributes (avoids a warning with recent Rustc).

---

This code needs significant optimisation (some enabled by GATs, some to do with filtering, probably some other opts too), but appears to work fine other than perf.

One notable possibility: `update_widgets` gets called whenever scrolling occurs, which may be multiple times per frame — especially since rendering appears to only happen after event handling is finished. With fast scrolling this can essentially block drawing until the mouse stops moving. We could try solving this by only calling `update_widgets` after receiving all events (via `ManagerState::update`), but in this case state would be incorrect if other event handlers referring to the list's children somehow get called before `update`.

### DIFF
--- a/kas-macros/src/lib.rs
+++ b/kas-macros/src/lib.rs
@@ -574,9 +574,9 @@ pub fn make_widget(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     // TODO: we should probably not rely on recursive macro expansion here!
     // (I.e. use direct code generation for Widget derivation, instead of derive.)
     let toks = (quote! { {
+        #[derive(Debug, kas::macros::Widget)]
         #handler
         #extra_attrs
-        #[derive(Debug, kas::macros::Widget)]
         struct AnonWidget #impl_generics #where_clause {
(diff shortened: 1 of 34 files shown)


dhardy:
Under further testing with release optimisations things appear sufficiently fast that drawing is not blocked even with rapid scrolling via the scrollbar, so it may be the `update_widgets` performance isn't a big issue.

Even with release opts, filtering is slow with >10'000 items, so optimisations are still desired for the filter (possibly only filtering entries during iteration/retrieval, though this requires we need to use an upper-bound when calling `len` or some such — possible since even the scrollbar size is independent of len above some threshold).

PR #167, opened 2021-02-24
Misc: programmatic selection, reduced draws, nav changes, ...

dhardy:
Handling for keys changes: `Escape` can now be sent to widgets in more cases, only deselecting if not otherwise used.

Several small changes and fixes; see below.


dhardy:
The Gallery's pop-up editor can now be closed with Alt+S or Alt+C, or (if no widget has nav focus) with the Return or Escape keys. But this is a bit weird: after editing, pressing Esc repeatedly will close (cancel), but accepting requires pressing Esc (to drop char focus), Esc (to drop nav focus), Return (to accept).

PR #179, opened 2021-03-31
CI: re-enable windows tests; add stable toolchain tests

dhardy:


dhardy:
Stable build fails: cycle detected:
```
error[E0391]: cycle detected when computing the bounds for type parameter `V`
  --> src/widget/view/single_view.rs:31:15
   |
31 |     V: Driver<T::Item> = driver::Default,
   |               ^^^^^^^
   |
note: ...which requires computing the bounds for type parameter `T`...
  --> src/widget/view/single_view.rs:30:38
   |
30 |     T: SingleData + UpdatableAll<(), V::Msg> + 'static,
   |                                      ^^^^^^
   = note: ...which again requires computing the bounds for type parameter `V`, completing the cycle
note: cycle used when computing the bounds for type parameter `V`
  --> src/widget/view/single_view.rs:31:5
   |
31 |     V: Driver<T::Item> = driver::Default,
   |     ^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0391`.
error: could not compile `kas`
```
I believe this affects all three view widgets. While it can be worked around by factoring out either `V::Msg` or `T::Item` as its own parameter, it appears that the next Rust release (current beta) can already support this, so I prefer to instead target 1.52.0.

PR #190, opened 2021-04-26
Rework text drawing: another atlas; replace wgpu_glyph with ab_glyph

dhardy:
Replace `wgpu_glyph` with our own texture atlas code. *Mostly* works with a couple of caveats:

- ~~a few glyphs have visual defects (why?)~~
- an `EditBox` inside a `ScrollRegion` doesn't clip text properly since `clip_region` is not currently recursive (will fix later since I plan to rework the other use of `clip_region`, for drawing menu pop-ups)

Also:

- some changes to atlases and images modules
- use alpha channel of colour for basic draw routines


dhardy:
Support sub-pixel precision, based on font size (height). To my eyes it appears slightly worse for normal-sized fonts but is definitely better for tiny fonts, hence it is currently only enabled for `height <= 20 px`, with the degree of sub-pixel precision determined by:
```rust
(30.0 / height).round().clamp(1.0, 15.0)
```
Eventually this should be run-time configurable but it's not yet.

Visual defects are gone (not sure why).

dhardy:
Notes on text bounds: previously, `wgpu_glyph` was calling `set_scissor_rect` for each text object drawn. Additionally we adjusted rectangles used for underline/highlight/edit marker. In this PR I had to make a choice:

### Maintain per-text-object scissor rects

This would have required further adaptations to the atlas code (possibly to the point that code-share with image rendering no longer made sense), and thus was not trivial to implement.

### Use DrawHandle::clip_region where required

`DrawHandle::clip_region` already exists for this purpose: setting up a new scissor region. Using it only where text-scrolling / overflow may be expected (currently only `EditField`) is easy; using it for *all* text drawing would not be.

Advantages:

- much easier to implement
- two fewer parameters to text methods (`bounds` and `offset` are dropped)
- allows removal of manual rect clipping in text-effects and edit-marker methods

Disadvantages:

- widget draw routine needs to use this explicitly where required for text clipping
- there might be a few corner cases where text overflows labels incorrectly, but nothing observed
- would be a slightly inefficient approach if used for to clip all text objects, but it is not

---

Ultimately I chose the latter option since it allows much cleaner code.

PR #193, opened 2021-05-01
Update to wgpu-rs v0.8

dhardy:
Most examples work. Mandlebrot does not.


dhardy:
Unfortunately committing this now breaks the mandlebrot demo, but this is preferable to waiting on a wgpu fix.

PR #197, opened 2021-05-10
Add dirty flag to config (limit writes)

dhardy:


dhardy:
I now wonder whether `config` / `serde` support should be optional due to extra code complexity. Well, it's done now.

PR #201, opened 2021-06-10
Fontdue rasterizer and other font adjustments

dhardy:
Several improvments to text rendering, plus font raster configuration (under theme configuration).

Enables usage of the https://github.com/mooman219/fontdue rasterizer. Subjectively it may be a little better but still isn't perfect and noticably slows down start-up time.

Another option is to align to the horizontal side bearing using `ab_glyph`. In some ways this is better than sub-pixel alignment, in some ways not. Either way it's not close to [grid fitting](https://docs.microsoft.com/en-us/typography/opentype/spec/ttch01#grid-fitting-a-glyph-outline).

The biggest improvement to rendering was one of the simplest changes: vertically aligning text lines to the pixel grid during layout (in `kas-text`). This is enabled by default; other options (including sub-pixel positioning) are now disabled by default which seems to be the best *general* solution.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
         run: cargo test --manifest-path kas-theme/Cargo.toml --all-features
       - name: test (kas-wgpu)
         run: |
-          cargo test --manifest-path kas-wgpu/Cargo.toml --no-default-features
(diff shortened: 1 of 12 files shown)


dhardy:
That test works for me (also on the latest nightly) and the error message is incomprehensible. I'll ignore for now.

PR #211, opened 2021-06-24
fix some clippy warnings

remilauzier:


src/layout/size_rules.rs
```diff
@@ -437,7 +437,6 @@ impl SizeRules {
             let target = target - margin_sum;
 
             if sum == target {
-                return;
```
dhardy, line 440: You tell me whether the method's flow is clear enough. May be okay, but definitely over-sized.

src/widget/slider.rs
```diff
@@ -200,9 +200,9 @@ impl<T: SliderType, D: Directional> Slider<T, D> {
             a = b - a;
         }
         let value = a + self.range.0;
-        let value = if !(value >= self.range.0) {
+        let value = if value < self.range.0 {
             self.range.0
-        } else if !(value <= self.range.1) {
+        } else if value > self.range.1 {
```
dhardy, line 205: These are not equivalent (consider NaN).

dhardy: Your solution here is not what [the lint](https://rust-lang.github.io/rust-clippy/master/index.html#neg_cmp_op_on_partial_ord) suggests — but that would be too verbose (impairing readability) here. I'm not entirely happy with just squelching the lint, but I guess for now it will do.

dhardy:
Re-running tests; not obvious why they failed.

PR #215, opened 2021-07-01
Add canvas widget

dhardy:
This is built over the [tiny-skia API](https://docs.rs/tiny-skia) (which is already used by the `Svg` widget).
The canvas has a built-in cache since it is rendered on the CPU then uploaded as a GPU texture.

Potentially we could add other canvas widgets over other draw APIs in the future, in which case this widget should be renamed.

Note there is already another option for drawing: constructing a new widget with code in the `Layout::draw` implementation (not cached); see e.g. `clock` or `mandlebrot` examples.

Also: some optimisations to text drawing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,9 @@ json = ["config", "serde_json"]
 # Enable support for RON (de)serialisation
 ron = ["config", "dep_ron"]
 
+# Support canvas widget
+canvas = ["tiny-skia"]
+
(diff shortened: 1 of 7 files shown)


dhardy:
More thoughts: using a different library like [lyon](https://crates.io/crates/lyon) we could have GPU-accelerated rendering for the canvas and potentially for SVGs too. Probably worth investigating at some point.

PR #226, opened 2021-07-30
Add DerivedDrawHandle

dhardy:
The intention here is to allow themes like `ShadedTheme` to be incomplete by providing some base implementation to fill in missing methods. Currently `ShadedTheme` does effectively the same thing but with stub functions; part of the point of this change is to make it non-breaking (or less breaking) to extend `DrawHandle` and `SizeHandle`.

**Blocker**:
```
error[E0119]: conflicting implementations of trait `draw::handle::DrawHandle` for type `std::boxed::Box<_>`
   --> src/draw/handle.rs:737:1
    |
568 | impl<D: DerivedDrawHandle> DrawHandle for D {
    | ------------------------------------------- first implementation here
...
737 | impl<H: DrawHandle> DrawHandle for Box<H> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `std::boxed::Box<_>`
    |
    = note: downstream crates may implement trait `draw::handle::DerivedDrawHandle` for type `std::boxed::Box<_>`
```
Fixing this would need some form of specialisation or negative trait bounds (e.g. rust-lang/rfcs/issues/1834).

An **alternative** would be to add the `base*` methods and default implementations to `DrawHandle` itself; this is less desirable since themes without a base (e.g. `FlatTheme`) would be forced to panic in the `base*` methods, resulting in run-time failure if the theme is incomplete (and only then on examples which use those size/draw methods).

**Incomplete** in this PR:

- switch `ShadedTheme` to impl `DerivedDrawHandle`
- add `DerivedSizeHandle`

### DIFF
--- a/src/draw/handle.rs
+++ b/src/draw/handle.rs
@@ -435,6 +435,238 @@ pub trait DrawHandleExt: DrawHandle {
 
 impl<D: DrawHandle + ?Sized> DrawHandleExt for D {}
 
+#[cfg_attr(not(feature = "internal_doc"), doc(hidden))]
+#[cfg_attr(doc_cfg, doc(cfg(internal_doc)))]
+pub trait DerivedDrawHandle {
+    type H: DrawHandle;
+    fn base(&self) -> &Self::H;
+    fn base_mut(&mut self) -> &mut Self::H;
+
+    /// Access a [`SizeHandle`]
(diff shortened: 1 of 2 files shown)


dhardy:
This is made obsolete by #327, though is potentially another alternative.

PR #234, opened 2021-08-06
Improve keyboard navigation

### LINKED ISSUE #231: Keyboard navigation is unintuitive
If I don't know which shortcuts I have to press (e.g. <kbd>Esc</kbd> to move focus away from a text input), this GUI framework is not usable with a keyboard. Please make sure it behaves the same as other GUI frameworks, so it can be used without reading a manual.

Here are some things I expect from a GUI:

* When a text input is focused, <kbd>Tab</kbd> focuses the next widget and <kbd>Shift+Tab</kbd> focuses the previous widget. There should not be the need to press <kbd>Esc</kbd>
* When pressing <kbd>Tab</kbd> and there is no next widget, focus the first widget. Pressing <kbd>Shift+Tab</kbd> when there is no previous widget focuses the last widget.
* Clicking (or pressing the left mouse button down) on a widget focuses the widget
* Activating a widget with a keyboard shortcut (e.g. <kbd>Alt+1</kbd> in the gallery) focuses the widget
* The menu bar usually isn't part of the tab order, it can only be focused with <kbd>Alt+...</kbd>
* When opening a menu in the menu bar and closing it again, focus is returned to the previously focused widget
* Pressing <kbd>Esc</kbd> doesn't remove focus from the active widget

Please consider adopting this behaviour.

dhardy:
Fixes most of #231 plus a few more things:

- Escape does not clear nav focus
- (Shift+)Tab navigates from widgets with char focus (i.e. edit box)
- Key event sending now tries multiple paths until one does *not* return `Response::Unhandled`: (1) the widget with char focus or nav focus
- Allows limited commands (deselect, copy) to be sent to a `ScrollLabel` with selection focus
- Set (keyboard) navigation focus on mouse click or touch (configurable)
- Allow widgets like `EditBox` to request char focus on nav focus
- Make `next_nav_focus` wrap after first/last available widget
- Fix closing pop-ups with <kbd>Esc</kbd>
- `EditGuard::focus_gained` can no longer send a message (breaking change)
- Remove `read_only` state of `Manager`, used while sending pending events (this potentially allows widgets to cause infinite loops, but was causing issues; probably it's okay to rely on correct widget code here)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,10 +19,13 @@ jobs:
       - name: Install dependencies
         run: sudo apt-get install -y libxkbcommon-dev libxcb-shape0-dev libxcb-xfixes0-dev
       - name: Build examples
-        run: cargo build --manifest-path kas-wgpu/Cargo.toml --release --example calculator --example gallery --example mandlebrot
+        run: |
+          cargo build --release --example calculator --example gallery
+          cargo build --release --manifest-path examples/mandlebrot/Cargo.toml
       - name: Prepare
         run: |
-          strip target/release/examples/calculator target/release/examples/gallery target/release/examples/mandlebrot
+          strip target/release/examples/calculator target/release/examples/gallery target/release/mandlebrot
+          mv target/release/mandlebrot target/release/examples/
           cp -a kas-wgpu/res target/release/examples/
       - name: Upload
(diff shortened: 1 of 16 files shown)


dhardy:
I have some ideas to fix menu-related navigation too, but will create a new PR for that.

Issue #238, opened 2021-08-11
Drawing shadows in themes


dhardy:
I was thinking some more about how to implement shadows (see #237):

Drawing offset, coloured frames for each widget is not too difficult. The problem is, where? Widgets are not supposed to draw in their outer margin area. *Usually* there's nothing else drawn there, but if the widget is but inside a container then that container is allowed to draw there.

Alternatively we could allocate extra area within the widgets proper, but then the shadow is a clickable part of the button.

The other possibility is to do a post-processing step: (1) add a depth buffer and possibly some type of mask, (2) draw shadows based on the depth and some light-direction. This allows some flexibility: soft shadows in all directions, user-configuration. It's also extra work, may be less portable and may have some other issues that need fixing.

hummingly:
Clickable shadows would indeed break the illusion. I do web dev mostly, so I expect box shadows to draw outside the boundary. It has the benefit of preserving the space.

While it will add more work, it will be necessary for implementing many designs, especially if someone wants to implement a platform specific one.

Now that I think of, popovers usually have a big drop shadow and then sometimes even in combination with background blur.

Oh, and most of all those acryl materials use the compositor too to get those see through blurred backgrounds.

One way or another, you can't really get away from a post-processing step.

hummingly:
The changes look nice. I actually had the shaded theme when I made themockup and had therefore a lot of many inner shadows. Do you plan to keep the shaded theme?

dhardy:
I wrote some custom shaders for the shaded theme to get those bevelled edges, so I'm not planning no just throwing it away! Possibly it would look good with shadows too (e.g. on the slider handle and buttons), though there may be some minor issues with draw order.

Your mock art used flat shading (other than the drop shadows) and simple line borders, hence it seemed closer to the old `FlatTheme`, which also needed some work anyway IMO.

Do you have further suggestions?

hummingly:
Yeah, the inner shadows were so thin that they ended up looking like borders. I will update the design to use borders directly. Shaded themes do not really work well with thin and small widgets as it seems. I wonder how a glow effect with shaded buttons would look. There is not so much going on in this design direction, so I would need to experiment a bit.

dhardy:
Ah. I did notice the inner shadow, then forgot about it... there's not much room before obscuring contents in e.g. a text box. It might be possible I guess, especially with high DPI.

[...]

PR #239, opened 2021-08-16
Simple shadows for FlatTheme

dhardy:
Implements basic shadows. CC #238, @hummingly.

Contrary to the discussion in #238, these are implemented using simple shaders, without even a mask (just relying on the draw order). Implications:

- Using custom size, offset and colours (glow effects) is easy
- Lower GPU load & memory usage, though this is probably irrelevant
- Shadows may look incorrect with too small margins (e.g. if a button is directly inside a frame, the frame will be drawn over the button's shadow)
- Sub-menus appear over parents instead of alongside with merged shadows
- To get outer curvature correct, shadows start inside the shape; with circles they start further inside than frames, hence the strongest visible part of the shadow is fainter on circles
- Outer curvature is currently squashed/stretched instead of simply offset; I could change this but not sure which is better.

I may leave shadows like this for now, and later implement post-processing only for pop-up menus — though those pop-ups are *supposed* to be rendered by the system compositor anyway (winit doesn't support this yet).

gallery-shadows

Forgot to show it in the video, but the under-bar on the edit box now has a small shadow too, but only on hover and focus.
edit_bar_shadow


dhardy:
I reduced the shadows a bit over the above:
 shadows

PR #241, opened 2021-08-19
WGPU 0.10, Image and Canvas scaling, tolerate missing config files

dhardy:
@kvark you'll be pleased to know that the GL backend "just works" with WGPU 0.10.0.

Unfortunately the Mandlebrot example now doesn't work at all with 64-bit shaders (leaves a black screen). It works fine in 32-bit mode (now the default for compatibility reasons). I don't see any useful warnings. Shall I try a renderdoc capture?

Other changes:

- `Image` now scales with the scale factor
- Adjusted scaling configuration of `Image` and `Canvas`
- Read and ReadWrite config modes now tolerate missing files
- Update to README


kvark:
Not sure what's up with 64-bit stuff. If you file a bug with an attached [API trace](https://github.com/gfx-rs/wgpu/wiki/Debugging-wgpu-Applications#tracing-infrastructure) (zipped), that would be enough for us to investigate.

PR #258, opened 2021-11-13
autoimpl; revise widget macros

dhardy:
This PR revisits how widgets are implemented, also adding the `autoimpl` attribute-macro. WIP.


dhardy:
There's still quite a bit to do for this PR:

- `make_widget` should not invoke `widget`
- eliminate `core_data` methods?
- use `quote_spanned` when generating impls etc
- implement `Default`?

Scratchboard for syntax ideas: https://gist.github.com/dhardy/97efb394311b8b9687c0d817e95654b9

dhardy:
This is part of a series of changes aimed at improving the way widgets are constructed. It is not finished, but does achieve:

- replace `derive(Widget)` with `widget! { ... }` macro
- autodetection of whether widget traits like `Layout` are implemented directly
- macro sugar: `impl Trait for Self` syntax — use of `Self` to imply struct generics and bounds, with optional extras (this may be brought to a stand-alone crate later)
- new macro: `autoimpl` — more capable variant of `derive` (also may be brought to a stand-alone crate later)
- fix required by recent nightly compilers

PR #262, opened 2021-12-19
Force wgpu to use raw-window-handle 0.3.4 for winit compatibility

nyanpasu64:
kas-wgpu depends on winit 0.26 which exposes raw-window-handle 0.4 types in its public API. As a result, wgpu (which exposes raw-window-handle 0.3 types in its public API) must be forced to use raw-window-handle 0.3.4, which is compatible with 0.4.x. Otherwise in projects with a stale Cargo.lock containing raw-window-handle <= 0.3.3 but latest kas-wgpu, you'll get a trait mismatch build error.

See https://github.com/kas-gui/kas/discussions/261.


dhardy:
Sorry this got left so long!

PR #270, opened 2022-01-25
Theme animations + misc

dhardy:
- Remove `TkAction::ANIMATE`
- Rename `edit_marker` → `text_cursor`
- Make `text_cursor` blink
- Add `DrawCtx`
- Events: in `PressStart`, `start_id` is now `Optional<..>`; `PressEnd` has new field: `success`
- Replace `EventMgr::request_grab` with `grab_press` and `grab_press_unique`

Misc. fixes:

- `FlatTheme::slider` highlight with reversed direction
- `kas_wgpu` timer handling
- Ensure redraw and set nav-focus when list/matrix view selection changes

Possible future extension: possibly add `WidgetId` to `DrawCtx` and `&EventMgr` to `DrawHandle`s, letting the theme calculate needed parts of `InputState` directly. *But* a few widgets currently mutate `InputState` before passing; we'd need an alternative.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,10 +59,10 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        toolchain: [stable]
+        toolchain: [beta] # TODO: stable
         include:
           - os: ubuntu-latest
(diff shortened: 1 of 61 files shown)


dhardy:
Note — MSRV bumped to 1.59, which should release in February. Until then, use the beta!

PR #274, opened 2022-01-31
SizeRules::solve_seq_total: fix to avoid negative widths

### LINKED ISSUE #273: reproducible panic in SizeRules::solve_seq_total
Steps to reproduce:
- open kas/examples/data-list.rs
- change line 140: `let list = IndexedList::new_with_direction(Direction::Down, entries);`
to `let list = IndexedList::new_with_direction(Direction::Right, entries);`
- start example with `cargo run --example data-list`
- set Number of rows to 100
- see the panic message: `thread 'main' panicked at 'assertion failed: out.iter().all(|w| *w >= 0)', ~/workspace/kas/crates/kas-core/src/layout/size_rules.rs:417:13`

dhardy:
Closes #273

### DIFF
--- a/crates/kas-core/src/layout/size_rules.rs
+++ b/crates/kas-core/src/layout/size_rules.rs
@@ -632,7 +632,7 @@ impl SizeRules {
                 }
             }
 
-            while excess > 0 {
+            while excess > 0 && largest > 0 {
                 let step = (excess / num_equal).min(largest - next_largest);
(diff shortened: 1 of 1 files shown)


dhardy:
We still have two failing tests: rustdoc nightly because of an internal compiler error (reported) and 1.59 because MSRV was bumped to a beta version in #270.

Issue #278, opened 2022-02-05
Menu behaviour


captchaKing:
I made some changes on my fork. May i PR for any of this or all of them?

1. Menu-items double highlighting (TkAction::REDRAW somehow not passed) 
https://user-images.githubusercontent.com/63655582/152634186-d4aeab5d-cc59-44bd-9066-10e9e8c8b82a.mp4
2. Grabbing bug
https://user-images.githubusercontent.com/63655582/152634346-4731ccc3-f61a-4912-9043-d6226d198ea3.mp4
3. Minor annoyance - when grabbing ends out of menu-item, menu closes (and that's how it's done in QT and GTK, but not in Chrome or Firefox):
https://user-images.githubusercontent.com/63655582/152634649-93ea3e96-e77f-4617-932a-b3be937c7baf.mp4
But consider user doing it fast, he may think that click actually happened on menu-item and something changed in the app.
4. Inactive delayed-menu's item
https://user-images.githubusercontent.com/63655582/152655399-e887b92e-910a-49d1-a792-8253a2579f7d.mp4
5. Menubar's root item sometimes suddenly reopens on close
https://user-images.githubusercontent.com/63655582/152655615-f60da48e-8543-4126-a100-169ba88a6882.mp4

captchaKing:
4. Inactive delayed-menu's item
https://user-images.githubusercontent.com/63655582/152655399-e887b92e-910a-49d1-a792-8253a2579f7d.mp4
5. Menubar's root item sometimes suddenly reopens on close
https://user-images.githubusercontent.com/63655582/152655615-f60da48e-8543-4126-a100-169ba88a6882.mp4

dhardy:
Sure, a PR is welcome.

Point 3 was deliberate (I mostly copied Qt behaviour), but you are right that it's not optimal.

Otherwise they just sound like bugs. I've had enough bugs with menu logic that I don't know if a better approach is needed.

PR #280, opened 2022-02-07
Fix MenuBar behaviour

captchaKing:
issue #278

### DIFF
--- a/crates/kas-widgets/src/menu/menubar.rs
+++ b/crates/kas-widgets/src/menu/menubar.rs
@@ -105,11 +105,7 @@ widget! {
                     coord,
                     ..
                 } => {
-                    if !mgr.set_grab_depress(source, cur_id.clone()) {
-                        // If redraw hasn't happened, force it to fix
-                        // highligting of previously selected menu-item
-                        mgr.send_action(TkAction::REDRAW);
-                    }
+                    mgr.set_grab_depress(source, cur_id.clone());
 
                     let id = match cur_id {
                         Some(x) => x,
--- a/crates/kas-widgets/src/menu/menubar.rs
+++ b/crates/kas-widgets/src/menu/menubar.rs
@@ -24,8 +24,6 @@ widget! {
         core: CoreData,
         #[widget]
         pub bar: IndexedList<D, SubMenu<D::Flipped, W>>,
-        // Open mode. Used to close with click on root only when previously open.
-        opening: bool,
         delayed_open: Option<WidgetId>,
     }
 
@@ -49,7 +47,6 @@ widget! {
             MenuBar {
                 core: Default::default(),
                 bar: IndexedList::new_with_direction(direction, menus),
-                opening: false,
                 delayed_open: None,
             }
         }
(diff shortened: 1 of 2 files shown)


crates/kas-widgets/src/menu/menubar.rs
```diff
@@ -75,28 +72,31 @@ widget! {
                 } => {
                     if start_id.as_ref().map(|id| self.is_ancestor_of(id)).unwrap_or(false) {
                         if source.is_primary() {
-                            mgr.grab_press_unique(self.id(), source, coord, None);
+                            let any_menu_open = self.bar.iter().any(|w| w.menu_is_open());
+                            let press_in_the_bar = self.rect().contains(coord);
+
+                            match press_in_the_bar {
```
dhardy, line 78: This can be reduced to `if !press_in_the_bar || !any_menu_open { .. }`

crates/kas-widgets/src/menu/menubar.rs
```diff
@@ -107,46 +107,45 @@ widget! {
...
-                                mgr.set_nav_focus(id.clone(), false);
-                                let delay = mgr.config().menu_delay();
-                                mgr.update_on_timer(delay, self.id(), id.as_u64());
-                                self.delayed_open = Some(id);
-                            }
-                        }
+                    if !mgr.set_grab_depress(source, cur_id.clone()) {
+                        // If redraw hasn't happened, force it to fix
+                        // highligting of previously selected menu-item
+                        mgr.send_action(TkAction::REDRAW);
```
dhardy, line 113: This doesn't sound right — if `set_grab_depress` returns `false`, the depressed widget hasn't changed. Menu entries are highlighted if either the `depress` or the `nav_focus` state is true. A change to `nav_focus` should also trigger a redraw.

I can reproduce the issue even with this extra redraw command, though in this case mouse movement must be stopped immediately after the mouse touches the new menu title, or a redraw happens.

My guess: the issue is that `SubMenu::draw` explicitly sets `DEPRESS` when `self.popup_id.is_some()` — sending of `Event::PopupRemoved` is delayed slightly.

dhardy (commented):
Thanks for the fixes. Mostly it works, but:

- As noted, the double-highlight isn't really fixed
- With a menu open, clicking one entry and holding the button, moving over another entry, then releasing, will select the entry under the mouse when the button is released. This matches Qt behaviour, whereas Firefox will do nothing. The latter seems preferable?

crates/kas-widgets/src/menu/menubar.rs
dhardy: Better fix (possibly handling `PopupRemoved` is redundant here; probably still better to do so):
```rust
// submenu.rs

        fn close_menu(&mut self, mgr: &mut EventMgr, restore_focus: bool) {
            if let Some(id) = self.popup_id {
                mgr.close_window(id, restore_focus);
                self.popup_id = None;
            }
        }

        fn handle(&mut self, mgr: &mut EventMgr, event: Event) -> Response<M> {
            match event {
                // ...
                Event::PopupRemoved(id) => {
                    if self.popup_id.as_ref() == Some(&id) {
                        self.popup_id = None;
                    }
                    Response::Used
                }
```
Or we could simply request a redraw in `PopupRemoved` (though this may redraw twice, flashing the double-highlight).

captchaKing:
> * With a menu open, clicking one entry and holding the button, moving over another entry, then releasing, will select the entry under the mouse when the button is released. This matches Qt behaviour, whereas Firefox will do nothing. The latter seems preferable?

I tested in Firefox 96, Electron (VSCodium), Qt 5.15.2, Gtk3 all of them select on releasing. It's your decision entirely. Should i fix?

dhardy:
> I tested in Firefox 96

I'm using the same version and it doesn't happen. Also, drag-clicking the menu-button then releasing on an item doesn't work for some reason. Same in Thunderbird; I guess it's some tweakable. Never mind, guess we'll leave this as-is.

> cargo fmt is not working for me on proc.macros

Inside the `widget!` and `make_widget!` macros it won't, no. Rustfmt *can* format macros, but only ones it recognises by name, so I plan to open a PR there (at least for `widget!`) at some point — but I may rename the macro first.

crates/kas-widgets/src/menu/menubar.rs
dhardy: Causing a redraw here should not be needed now.

dhardy (approved):

PR #291, opened 2022-03-04
update deps, add widget_index!, improve examples/markdown

dhardy:
- Update dependencies (minor changes)
- `examples/markdown.rs`: remove button and update on typing
- add `widget_index!` macro for use within `widget!` macro

### DIFF
--- a/crates/kas-macros/src/widget_index.rs
+++ b/crates/kas-macros/src/widget_index.rs
@@ -1,0 +1,98 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License in the LICENSE-APACHE file or at:
+//     https://www.apache.org/licenses/LICENSE-2.0
+
+use crate::args::{self, Widget};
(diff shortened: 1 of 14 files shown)


crates/kas-macros/src/widget_index.rs
```diff
@@ -0,0 +1,98 @@
...
+
+struct Visitor<'a> {
+    children: &'a [args::Child],
+}
+impl<'a> VisitMut for Visitor<'a> {
+    fn visit_macro_mut(&mut self, node: &mut syn::Macro) {
+        // HACK: we cannot expand the macro here since we do not have an Expr
+        // to replace. Instead we can only modify the macro's tokens.
+        // WARNING: if the macro's tokens are modified before printing an error
+        // message is emitted then the span of that error message is incorrect.
```
dhardy, line 63: Yes, this is a bit hacky, but still works well enough. The only alternative (as far as I am aware) would be to match one level up, which would require matching both expressions and patterns (and maybe more) to catch everywhere the macro might be used.

PR #315, opened 2022-05-05
Trait revision part 4: remove Draw::recurse; 'component' from syntax; Widget: Any

dhardy:
Following up on #314, we remove `Draw::recurse` (instead, manually setting the `id` where required). The first change is not great since now every `fn draw` with local drawing needs an extra line to set the `WidgetId` used (otherwise highlighting will be incorrect).

Removing `Draw::recurse` allows removal of the need for the `component` keyword in layout syntax.

### Alternative: keep `component` syntax (or similar)

This works...

The aim was to unify widgets and components, but despite now using the same trait this is not achieved.

### Alternative: separate methods for Widget / Layout draw

We can have `Widget::draw` call `draw.set_id(self.id())` then call `Layout::draw`. If the user only overrides `Layout::draw`, there's less code. (The same would be useful for `set_rect` and `find_id` methods, and also simplify macro logic.)

Problem: a widget now has two `draw` methods, thus calling `draw` is ambiguous.

**Solution 1:** use different method names for `Widget` vs `Layout`. But this is unnecessarily confusing towards users? And, if an unknowing user calls the wrong method, will result in some subtle issues?

**Solution 2:** use some type of wrapper (like Druid's `WidgetPod`), and move the second `draw` (etc.) to that. *But* custom widgets embed other widgets all the time; requiring a wrapper here is a nuisance.

**Solution 3:** use a widget wrapper (solution 2), but have the `#[widget]` macro automatically wrap any child which is a `#[widget]`. Implement `Deref` and `DerefMut` on this child so that *most* interactions with it give the illusion that there is no wrapper. Caveat: constructors must acknowledge the wrapper (unless the macro gets *quite* hacky). Caveat: it's surprising behaviour (especially since I'd like to remove the `#[widget]` attribute currently required on children).

---

Other changes:

- Remove `Widget: Any` (and `'static` bound). This allows `&'a mut W` to implement `Widget` where `W: Widget`.
- Prefer to use `&dyn Widget` over `&W where W: Widget` in event manager code: dynamic dispatch is fine here.


crates/kas-resvg/src/svg.rs
```diff
@@ -151,6 +151,7 @@ impl_scope! {
 
         fn draw(&mut self, mut draw: DrawMgr) {
             if let Some(id) = self.image.as_ref().map(|h| h.id()) {
+                draw.set_id(self.id());
                 draw.image(self.rect(), id);
             }
         }
```
dhardy, line 157: This extra line is what I don't like about the change: `draw` methods must call `set_id` to allow drawing to respond to events (such as hover).

(In this case it's not even necessary since `draw.image(..)` doesn't depend on any event state, but I don't expect users to remember which draw calls *do* depend on event state.)

(And, yes, the "obvious" solution is to pass `self.id` into draw calls, where required. *But* components don't have a `WidgetId` to pass. A few PRs back, `Component::draw` had an extra parameter: `id: &WidgetId`. In order to unify components and widgets, this had to be removed (or added to widgets, and calls to `Widget::draw` updated to use it — which is close to what `Draw::recurse` did).

PR #316, opened 2022-05-09
Components are widgets

dhardy:
Alternative to #315: components are widgets. This resolves ambiguities over whether something is a widget or a component. It is sometimes a little inefficient (storing a `WidgetId` for things like a label which don't need one), but this is not significant.

- Replace `Layout` and `Mark` components with real widgets (removes `kas::component` module)
- Allow widgets introduced by layout syntax to be enumerated by `WidgetChildren` and configured
- Allow embedding of widgets within layout syntax (currently these are boxed; given something like RFC 2524 they could be placed inline, but again this inefficiency is not important)
- Remove support for glob (`*`) from layout syntax; this is less important (since widgets may be embedded) and potentially confusing given that layout syntax also declares child widgets

Example: the stopwatch can now be declared as follows, with the buttons inlined into the layout:
```rust
        #[widget{
            layout = row: [
                self.display,
                TextButton::new_msg("&reset", MsgReset),
                TextButton::new_msg("&start / &stop", MsgStart),
            ];
        }]
        struct {
            #[widget] display: impl HasString = Frame::new(Label::new("0.000".to_string())),
            saved: Duration = Duration::default(),
            start: Option<Instant> = None,
        }
```
Now, child widgets must appear as `#[widget]` fields in one of two cases:

- The field is directly read/written to by event handlers
- Construction of the field depends on constructor parameters or moved values

---

Inherits some other commits from #315: removing `Widget: 'static` bound; preferring `dyn Widget` over `W: Widget` in event-management code; some clean-up. Closes #315.

The `Label` widget now allows replacing the text class.

Minor change to layout of `examples/data-list*`


crates/kas-widgets/src/mark.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+    impl Layout for Self {
+        fn size_rules(&mut self, mgr: SizeMgr, axis: AxisInfo) -> SizeRules {
+            mgr.mark(self.style, axis)
+        }
+
+        fn draw(&mut self, mut draw: DrawMgr) {
+            draw.mark(self.core.rect, self.style);
+        }
+    }
```
dhardy, line 50: This inherits default implementations:

- `set_rect` just assigns to `self.core.rect` (fine)
- `find_id` returns `self.id()` if `self.core.rect` contains `coord` — okay, but since this widget *never* handles click events (or any events), maybe it's better to return `None` — that way, "forgetting" to use `#[widget]` on a field to configure the widget is harmless

The same applies to `Label` too. Not a change for this PR, but possibly a future one.

PR #324, opened 2022-06-03
Invisible scrollbars; mouse-hover notifications; misc

dhardy:
- New events: `LostNavFocus`, `MouseHover`, `LostMouseHover`
- Remove `Widget::hover_highlight` and `Widget::cursor_icon` methods (but the same properties on `#[widget]` remain)
- Add `Widget::pre_handle_event` hidden method to contain `hover_highlight` and `cursor_icon` macro-generated code
- Fix order of handling pending events (FIFO, not FILO)
- New `HasScrollBars` trait, because I'm too lazy to duplicate methdos over `ScrollBars` and `ScrollBarRegion`

This leads to one new feature: invisible scrollbars (visible on hover when range != 0). Tested with mouse only (should work with touch, but without hover effects it's not obvious they're there).


dhardy:
Added: show scrollbars when scrolled by dragging content (including momentum scrolling + delay). This makes them obvious enough with touch interaction.

Really, we should maybe copy Firefox's behaviour: a very tiny visible scrollbar when not focussed, and more visible one when interacted with. This is mostly about themes and theme API since the `ScrollBar` widget has all required state available when drawing.

PR #335, opened 2022-07-15
Updates; kas_view; 

dhardy:
Dependencies updated: `ron`, `usvg`, `resvg`.

Dependency removed: `stack_dst`. `Box<dyn ..>` is a simpler alternative and there are no obvious performance issues, so I conclude that use of `stack_dst` was a premature optimization. (Also quite `unsafe`!)

`#[widget { derive = self.foo; }]` now supports partial trait implementations.

Bump MSRV to 1.62.0.

`examples/sync-counter.rs`: use only a `Spinner` widget (simpler code).

Revise drivers:

- Rename `DefaultView` → `View`, `DefaultNav` → `NavView`
- Move `event::Config` impl to `driver::EventConfig`

Use Cargo features stabilized in 1.60.0:

- Weak dependencies: `serde` no longer implies `winit`
- Namespaced dependencies: rename `dep_ron` to `ron` using `dep:ron`
- Namespaced dependencies: make features explicit, not implicit from optional dependencies
- Replace `canvas` and `svg` features with `tiny-skia` and `resvg`, the latter including the `Canvas` widget; adjust how the import looks in `kas` doc

Crates:

- Bump all versions to 0.11
- New `kas_view` crate from the contents of `kas::widgets::view`; this is now exported as `kas::view`
- Revise doc for model/view


dhardy:
CI failed on nightly due to: https://github.com/NotIntMan/inplace_it/issues/10

PR #336, opened 2022-07-16
Work2

### LINKED ISSUE #277: Text clipping
Is it possible for Text to cut instead of word-wrapping? I'm trying to build Labels-holding Separator for header of [listview](https://i.stack.imgur.com/2FNbr.jpg), but when i collapse headers text always overlaps.
I must say i modified Separator, so its handles could collapse to some smaller size. And changed Label's TextClass to Button, so it wont word-wrap.

dhardy:
Fix #277: clip text to available rect when drawing.

Add some consts to `Vec2`, `Quad`.

Revise accelerator keys for the Gallery.

### DIFF
--- a/crates/kas-core/src/core/widget.rs
+++ b/crates/kas-core/src/core/widget.rs
@@ -348,7 +348,7 @@ pub trait Layout {
 ///         }
 ///
 ///         fn draw(&mut self, mut draw: DrawMgr) {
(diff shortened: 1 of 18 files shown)


dhardy:
This has a bug: it doesn't consider the scroll offset. Scrolled content is clipped to its unscrolled position.

PR #340, opened 2022-07-27
Update WGPU to v0.13

dhardy:
Replaces #331.

One shader was translated to WGSL as a proof of concept. GLSL shaders will remain to ease any future OpenGL implementation; in general there doesn't seem to be motivation to switch the shader language.


dhardy:
There *is* a motivation to switch to WGSL: no need for a GLSL compiler and the option to embed WGSL in Rust. I translated Mandlebrot's shaders [in this branch](https://github.com/kas-gui/kas/commits/mandlebrot-wgsl).

But, I feel, there are significant drawbacks:

- GLSL code is much cleaner
- The WGSL spec is incomplete (e.g. no mention of `f64` support yet, which does work, though literals appear to be missing)
- WGSL validation for embedded shaders only happens at run-time (though naga can be run directly for fast validation)

PR #343, opened 2022-08-08
Derive with children

dhardy:
An alternative explored while writing #342: extend derive mode, allowing child widgets with derive.

(Or, as an alternative to this, let the wrapping widget keep its own `id` but still forward events from `handle_event` to the inner widget.)

Problem with both of these approaches: when should `handle_scroll` get called?

-   If `handle_scroll` gets called after `handle_event` on self, then the method
    call must also be forwarded, which is a problem for `EditBox` (which needs to react to the inner `EditField` scrolling).
-   If not, and the inner `handle_event` causes a scroll, then `handle_scroll`
    would be called on the outer widget only if the inner widget was the
    direct event target. Workaround: the forwarding of `handle_event` must
    call `handle_scroll` itself? Possible, but over-complex.

Conclusion: neither of these approaches should be used.

### DIFF
--- a/crates/kas-core/src/core/widget.rs
+++ b/crates/kas-core/src/core/widget.rs
@@ -117,6 +117,12 @@ pub trait WidgetChildren: WidgetCore {
     }
 }
 
+/// Marker trait indicating that a widget has a fixed number of children
+#[autoimpl(for<T: trait + ?Sized> &'_ mut T, Box<T>)]
+pub trait WidgetChildrenConst: WidgetChildren {
+    const NUM_CHILDREN: usize;
+}
+
 /// Positioning and drawing routines for [`Widget`]s
 ///
 /// This trait is related to [`Widget`], but may be used independently.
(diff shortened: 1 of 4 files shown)


dhardy:
Further notes regarding changes to derive behaviour:

-   Interactive widgets must have a `WidgetId`
-   Parents do need a `WidgetId` simply in order to make message routing work
-   "Dumb" widgets like `Filler` and `Label` don't need a `WidgetId`, but
    a parent wishing to handle unhandled events on them would need them to have one.
    (For this reason, widgets only optionally having a `WidgetId` doesn't make sense.)
-   Derive mode widgets don't need a `WidgetId`: they directly refer to their
    derive target. This implies that `handle_event` should not be overridden
    (technically possible, but it should definitely forward events).
-   Widgets like `CheckButton`, `EditBox`: it's desirable to make them act like
    the inner interactive widget so that a parent can easily send them messages:
    in other words, derive mode is desired to let the outer widget masquerade as
    the input handler. Or is this actually important?
    Technically, we only need that the wrapping widget forwards events.

---

Closing: it was decided not to expand "derive mode" to allow children due to complexity (and especially due to possible confusion in how event-handling is affected).

PR #345, opened 2022-08-09
Rename Widget::spatial_nav → nav_next and auto-implement

dhardy:


dhardy:
Add:

- `non_navigable` modifier to `layout = ..;` property of `#[widget]`
- Autogenerate `Widget::nav_next` from `layout = ..;` property

Renames:

- `Widget::key_nav` → `Widget::navigable` — this is not only *keyboard* navigation since focus can be set via mouse click
- `Widget::spatial_nav` → `Widget::nav_next` — could have been `nav_order`, but 

Fix:

- Do not trigger accelerator keys with Ctrl/Shift/Logo modifier

Doc:

- Improve doc for `layout = ..;` property of `#[widget]`

PR #346, opened 2022-08-10
Revise shared-data: new RadioGroup type, use UpdateId less, add SharedArc

dhardy:


dhardy:
Breaking change: `Driver::set` method now takes a data item but not a reference to the shared data object. Rationale:

- fixes an issue where `List/MatrixView` could not know whether `set` returned `TkActions::empty()` because the widget needed no changes or because there was no data item
- *all* existing impls of `set` used `SharedData::get_cloned(key)` first to retrieve the item; this change simplifies impls without any foreseen limitations

PR #373, opened 2023-01-03
Pop-ups use ancestors' coordinate translations

dhardy:
Fixes bugs with pop-up menus inside scrolled regions


dhardy:
Clippy fails on nightly without an error message. Works for me (exact same rustc version and commit).

PR #379, opened 2023-01-31
Update winit, wgpu, dark-light, window_clipboard

dhardy:
Update to use the pending winit release, wgpu 0.15 and dark-light 1.0.

Uses https://github.com/hecrj/window_clipboard/pull/20.


dhardy:
Unfortunate that there isn't an official release of `window_clipboard` compatible with winit 0.28, but this will do for now.

PR #393, opened 2023-06-22
Widget traits: direct impl of Widget for ListView, move nav_next to Layout, re-order

dhardy:
Another small follow-up on #391, #392. This *should* be the last before making controversial changes (associated input data).

The most significant change here is that `ListView` and `MatrixView` now implement the `Widget` trait directly. *This will not be supported outside of Kas libraries.* Motivation: allow everything else to use a simpler variant of `fn nav_next`, which can also now be moved to the `Layout` trait. (We could, in the future, provide an alternative which allows the extra functionality required by `ListView` without use of hidden internals, but I'm not convinced of the need.)


dhardy:
Bonus content: turn `RootWidget` into a `Window`, replacing `trait Window`.

Issue #415, opened 2023-10-18
A Label with too much text (font and font size dependent) causes wgpu panic


m-hugo:
Displaying a big string in a scrolling view is a common use case but ScrollLabel::new(LOTS_of_TEXT) panics

error (on master but reproduces on alpha and 0.13):
```
thread 'main' panicked at 'Error in Surface::configure: Validation Error

Caused by:
    Not enough memory left
', /.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.17.1/src/backend/direct.rs:771:18
```

code
```rust
fn main() -> kas::shell::Result<()> {
    let theme = kas::theme::FlatTheme::new().with_font_size(15.0);
    kas::shell::DefaultShell::new((), theme)?
        .with(Window::new(ScrollLabel::new(LOTS_of_TEXT), "MCVE"))
        .run()
}
```
`Scroll<Label>` and just `Label` behave identically

for LOTS_of_TEXT you can use for example
```rust
r##"source html from https://github.com/kas-gui/kas"##
```
which displays at `.with_font_size(1.0)` but crashes at 2.0, an ebook chapter crashed at 14.0

dhardy:
I tried pasting a few markdown docs into the text editor of `examples/gallery.rs` and didn't see this error.

> Not enough memory left

This could be system-specific — one of the big issues with using hardware-accelerated rendering. I tried measuring usage locally and saw maybe 60-200MB (on top of approx. 2.4GB just from desktop apps).

Can you check your video RAM size and usage?

There should be a software-rendering fallback, but there isn't yet: #287

dhardy:
That's basically what I have (but only 8GB version), so probably something else...

I reproduced with that document. It turns out to only be one long (>16k) line that's responsible. Then I see this in the log:
```
[2023-10-19T13:29:37Z DEBUG kas_core::layout::sizer] find_constraints: min=Size(140, 107), ideal=Size(228, 10939), margins=Margins { horiz: (0, 0), vert: (0, 0) }
[2023-10-19T13:29:37Z INFO  kas_core::shell::window] new: constructed with physical size Size(342, 16409), scale factor 1.5
```
That's a stupidly big window, but the buffer should only be around 22MB. It might exceed the maximum texture size though.

---

The winit window builder:
```
[...]

PR #419, opened 2023-11-22
ScrollText widget, font sizes in pixels, AdaptWidget::on_messages, misc fixes

dhardy:
Most notable is probably the `ScrollText` widget, requested in #415 (CC @m-hugo). Not currently used.

Font sizes probably shouldn't use old imperial units intended for publishing on paper (even if Windows and many Linux DEs still do).

### DIFF
--- a/crates/kas-core/src/shell/mod.rs
+++ b/crates/kas-core/src/shell/mod.rs
@@ -53,6 +53,7 @@ enum ProxyAction {
 #[cfg(test)]
 mod test {
     use super::*;
+    use raw_window_handle as raw;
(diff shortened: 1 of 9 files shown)


m-hugo:
thanks, ScrollText works very well, I even understood how you fused Text and ScrollLabel by looking at this code but it involves a crazy amount of copy-pasting, can you use some of your macro magic to make a more composable widget system ?

dhardy:
Something to think about for sure. But pushing more complexity into macros isn't necessarily a good answer; you already accused my macros of being black magic!

PR #443, opened 2024-02-22
Add CellCollection for use by Grid

dhardy:
- Rename `GridChildInfo` → `GridCellInfo`
- Add `trait CellCollection`, macro `cell_collection!`
- Some fixes

### DIFF
--- a/crates/kas-core/src/core/collection.rs
+++ b/crates/kas-core/src/core/collection.rs
@@ -5,7 +5,9 @@
 
 //! The [`Collection`] trait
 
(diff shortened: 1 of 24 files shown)


dhardy:
This fits with other recent changes. Unfortunately, this particular change doesn't achieve much:

- It enables usage of `kas::widgets::Grid` with varying child widget types, however in this case cells cannot be added/removed and thus the existing `grid!` macro could be used
- The `grid!` macro could, when used as a stand-alone macro, be replaced by `Grid::new(cell_collection![..])`. However, when used as part of the layout of a custom widget, it cannot be.

PR #450, opened 2024-05-03
Update winit, wgpu, resvg

dhardy:
Regression: this always reports `Platform::Wayland` on Linux (https://github.com/rust-windowing/winit/pull/3672). This may affect decorations on X11, but it's not worth blocking the PR or finding a workaround for now.

Migrates to winit's new `ApplicationHandler` (straightforward).

### DIFF
--- a/crates/kas-core/Cargo.toml
+++ b/crates/kas-core/Cargo.toml
@@ -121,7 +121,7 @@ version = "0.5.0" # used in doc links
 
 [dependencies.winit]
 # Provides translations for several winit types
-version = "0.29.2"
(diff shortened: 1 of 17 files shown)


dhardy:
Also revises feature categories:

- `nightly` now implies `stable` too
- Addition of `experimental`
- Demotion of `x11` from `minimal` to `stable`

Feature `harfbuzz` is removed (it can still be enabled with a direct dependency on `kas-text`, but as far as I am aware there is no reason to use this over `shaping`).

Feature `dynamic` is now excluded from all testing (less failures please!).

PR #477, opened 2025-02-16
Better handling for close/suspend/exit

dhardy:
Ctrl+Q was already mapped to `Command::Exit`; this is now handled.

Similarly, Alt+F4 maps to `Command::Close` which is now handled: close the popup or window if none.

Also added `AppData::suspended` to allow data saving on suspend/exit.


dhardy:
Also: update wgpu to 24.0.1

PR #489, opened 2025-04-06
Update MSRV to 1.86 for dyn trait upcast; update deps

dhardy:
Trait upcasting allows the pre-formatted part of `Erased` messages to be removed.

I also investigated removing `fn Tile::as_tile(&self) -> &dyn Tile`. This almost worked:

- `fn TileExt::find_widget` is implemented for all `T: Tile` and must be able to convert `&self` to `&dyn Tile`; it appears we need `T: std::marker::Unsize<dyn Tile>` (unstable) to make this work
- While some usages of `w.as_tile()` could be replaced with just `&w` (or just `w`), some required `w as &dyn Tile` or similar for unclear reasons; overall code looks worse as a result

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,7 +71,7 @@ jobs:
         toolchain: [stable]
         include:
           - os: ubuntu-latest
-            toolchain: "1.81.0"
+            toolchain: "1.86.0"
             variant: MSRV
           - os: ubuntu-latest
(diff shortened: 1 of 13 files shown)


dhardy:
- Don't update `smol_str` (this is waiting on `winit`)
- Update `Cargo.lock.msrv`
- Clippy fixes

PR #491, opened 2025-04-23
Revise mouse & touch code; add click-click-drag gesture

dhardy:
New `mouse` and `touch` modules and some code re-organisation.

More `ScrollDelta` methods for easier `Event::Scroll` handling.

Change touchpad-scroll handling in Mandlebrot example: pan by default, zoom with Ctrl key.

Add click-click-drag gesture, used to enable scale and rotate "panning" by mouse: click to set a fixed point, then click+drag to move a second point. Draw a marker for the fixed point.

Fix `ComboBox` label's initial/update value.

Partial fix for frame-updates while panning getting "stuck"... this issue now mostly doesn't occur, except when it does (which never happens when I insert extra debug instrumentation code, but otherwise sometimes does on my Intel/Wayland/KDE laptop). Probably I should just take the easy fix here (force re-drawing while a pan event is in progress, as done in 16a7378). See also https://github.com/rust-windowing/winit/issues/4202.


crates/kas-core/src/event/events.rs
```diff
@@ -673,7 +673,54 @@ impl FocusSource {
...
+    }
+
+    /// Convert to a zoom factor
+    pub fn as_factor(self, _: &EventState) -> f64 {
+        // TODO: this should be configurable?
+        match self {
+            ScrollDelta::Lines(_, y) => -0.5 * y as f64,
+            ScrollDelta::Pixels(Offset(_, y)) => -0.01 * y as f64,
+        }
+    }
```
dhardy, line 725: Ideally both constants would be configurable, but I'll leave this for later.

crates/kas-core/src/event/events.rs
```diff
@@ -723,4 +723,31 @@ impl ScrollDelta {
...
+    /// though both are supported (for example, a map view where click-and-drag
+    /// may also be used to pan). Mouse wheel actions always zoom while the
+    /// touchpad scrolling may cause either effect.
+    pub fn as_factor_or_offset(self, cx: &EventState) -> Result<f64, Offset> {
+        if matches!(self, ScrollDelta::Lines(_, _)) || cx.modifiers().control_key() {
+            Ok(self.as_factor(cx))
+        } else {
+            Err(self.as_offset(cx))
+        }
+    }
```
dhardy, line 752: Possibly the switching behaviour should be configurable, but for now I don't have a justification for that.

PR #500, opened 2025-06-02
Update kas-text: use Fontique + Swash

dhardy:
Updates kas-text for https://github.com/kas-gui/kas-text/pull/94

This implements font resolution by script, thus many more characters resolve to a valid glyph, though not all.

Some functionality is lost here:

- Read/write support for font configuration: the selector changed, is non-trivial to convert due to usage of a lookup table. TODO for later.
- `SizeCx::line_height` was removed since this depends on the font, which now depends on the text script. Instead we specify `EditField` height in `Em`s (or actual text height, whichever is greater).

### DIFF
--- a/crates/kas-core/src/theme/text.rs
+++ b/crates/kas-core/src/theme/text.rs
@@ -11,18 +11,17 @@ use crate::cast::Cast;
 #[allow(unused)] use crate::event::ConfigCx;
 use crate::geom::{Rect, Size};
 use crate::layout::{AlignHints, AxisInfo, SizeRules};
-use crate::text::fonts::{FontId, InvalidFontId};
+use crate::text::fonts::{FontSelector, InvalidFontId};
 use crate::text::format::{EditableText, FormattableText};
 use crate::text::*;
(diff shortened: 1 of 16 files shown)


crates/kas-core/src/theme/text.rs
```diff
@@ -397,101 +396,62 @@ impl<T: FormattableText> Text<T> {
...
-            Status::Configured => self
+            Status::New => self
                 .display
-                .prepare_runs(&self.text, self.direction, self.font_id, self.dpem)
-                .map_err(|_| {
-                    debug_assert!(false, "font_id should be validated by configure");
-                    NotReady
-                })?,
+                .prepare_runs(&self.text, self.direction, self.font, self.dpem)
+                .expect("no suitable font found"),
```
dhardy, line 404: `prepare_runs` may now return `Err(NoFontMatch)`. I opted to handle this via panic here since it *shouldn't happen* (there should always be some fallback font/glyph, even if it's just the "tofu" box).

PR #506, opened 2025-06-12
Update impl-tools: use `#[impl_self]`

dhardy:
Update to `impl-tools-lib` v0.11.2 and use the new `#[impl_self]` macro instead of `impl_scope!` where possible. This allows usage of `cargo fmt` within widget definitions.

See: https://github.com/kas-gui/impl-tools/pull/53

Several usages of `impl_scope!` remain to support `#[impl_default]`. These should be replaced using RFC 3681 (default field values) later.

### DIFF
--- a/crates/kas-widgets/src/stack.rs
+++ b/crates/kas-widgets/src/stack.rs
@@ -29,7 +29,8 @@ impl State {
     }
 }
 
-impl_scope! {
+#[impl_self]
+mod Stack {
     /// A stack of widgets
     ///
     /// A stack consists a set of child widgets, "pages", all of equal size.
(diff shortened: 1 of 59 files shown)


crates/kas-widgets/src/stack.rs
```diff
@@ -42,19 +43,32 @@ impl_scope! {
...
+                align_hints: AlignHints::NONE,
+                widgets: Vec::new(),
+                active: 0,
+                size_limit: usize::MAX,
+                next: 0,
+                id_map: HashMap::new(),
+            }
+        }
+    }
+
```
dhardy, line 71: This is the only case where I opted to replace `#[impl_default]` with a hand-written `Default` impl. Why:

1. This is a widget definition with a significant amount of code, therefore more important than most other uses of `#[impl_default]` to be automatically formattable.
2. `#[derive(Default)]` would use the wrong generic bounds here, so RFC 3681 is not sufficient to replace this usage of `#[impl_default]`

PR #507, opened 2025-06-12
Revise widget macros

dhardy:
"Derive mode" is split out into a dedicated attribute macro, `#[derive_widget]`. Both this and `#[widget]` have revised argument handling, with a new `#[layout]` attribute.

Experimentally, this adds support for using `#[widget = expr]` syntax to map input data.

# Motivation

`#[widget]` and `#[derive_widget]` already functioned very differently before this change; giving `#[derive_widget]` its own attribute improves documentation and discoverability.

Other changes move macro arguments to trait items which can be better documented.


crates/kas-widgets/src/edit.rs
```diff
@@ -734,11 +734,7 @@ mod EditField {
     /// scratch on each key stroke). Regardless, this approach is not designed
     /// to scale to handle large documents via a single `EditField` widget.
     #[autoimpl(Clone, Debug where G: trait)]
-    #[widget{
-        navigable = true;
-        hover_highlight = true;
-        cursor_icon = CursorIcon::Text;
-    }]
+    #[widget]
     pub struct EditField<G: EditGuard = DefaultGuard<()>> {
```
dhardy, line 738: Macro args are replaced ...

crates/kas-widgets/src/edit.rs
```diff
@@ -835,8 +831,16 @@ mod EditField {
...
+        const REDRAW_ON_HOVER: bool = true;
+        const NAVIGABLE: bool = true;
+
         type Data = G::Data;
 
+        #[inline]
+        fn hover_icon(&self) -> Option<CursorIcon> {
+            Some(CursorIcon::Text)
+        }
+
```
dhardy, line 843: ... with assoc. items.

crates/kas-widgets/src/frame.rs
```diff
@@ -52,10 +52,8 @@ mod Frame {
     // NOTE: this would use derive mode if that supported custom layout syntax,
     // but it does not. This would allow us to implement Deref to self.inner.
     #[derive(Clone, Default)]
-    #[widget{
-        Data = W::Data;
-        layout = frame!(self.inner).with_style(self.style);
-    }]
+    #[widget(type Data = W::Data)]
+    #[layout(frame!(self.inner).with_style(self.style))]
     pub struct Frame<W: Widget> {
```
dhardy, line 57: `Data` and `layout` arg changes

dhardy:
> Experimentally, this adds support for using `#[widget = expr]` syntax to map input data.

Attributes of the form `#` `[` _SimplePath_ `=` _Expr_ `]` are allowed by the [rust reference](https://doc.rust-lang.org/reference/attributes.html#meta-item-attribute-syntax) and by `syn::Meta` and do work, according to tests. Despite this, if the attribute is not matched, an additional error is shown:
```
error: attribute value must be a literal
   --> crates/kas-widgets/src/check_box.rs:186:20
    |
186 |         #[widget = &()]
    |                    ^^^

error: cannot find attribute `widget` in this scope
   --> crates/kas-widgets/src/check_box.rs:186:11
    |
186 |         #[widget = &()]
    |           ^^^^^^
```

PR #516, opened 2025-07-04
Revise Spinner and Slider traits; more standard messages

dhardy:
This supports external programmatic control of input widgets.

Significant:

- Remove `fn EventCx::send_command`; instead `fn EventState::send` has special handling for `Command` and `ScrollDelta`
- More widget support for `Activate`
- New standard messages `SetValueF64`, `SetValueString`, `SetIndex`; widget support for these along
- Standardised documentation of supported messages (still just doc)
- `MessageStack` moved to mod `runner`
- Rename `Spinner` → `SpinBox`
- Add `fn SpinBox::with_unit` to display a "unit" label within the box
- `trait SliderValue` supports cast to and from `f64` and drops support for `Duration`
- `trait SpinValue` supports cast to and from `f64` and has further revisions


dhardy:
Some `SpinBox`es:
image

PR #517, opened 2025-07-07
#[widget] macro revisions

dhardy:
New `#[collection]` attribute to replace explicit `num_children` / `get_child` / `child_node` definitions on a few widgets.

Drop support for `type Data = ...` in `#[widget]` attr args; instead:

- Assume `Data = ()` without `impl Events { ... }`, `impl Widget { ... }` or children
- Assume `Data = <#field_ty as ::kas::Collection>::Data` without `impl Events { ... }`, `impl Widget { ... }` or children but with `#[collection]` attribute on some field
- Require explicit definition (on `Events` or `Widget`) otherwise

### DIFF
--- a/crates/kas-widgets/src/filler.rs
+++ b/crates/kas-widgets/src/filler.rs
@@ -14,7 +14,7 @@ mod Filler {
     /// This widget has zero minimum size but can expand according to the given
     /// stretch priority.
     #[derive(Clone, Debug, Default)]
-    #[widget(type Data = ())]
+    #[widget]
     pub struct Filler {
         core: widget_core!(),
         horiz: Stretch,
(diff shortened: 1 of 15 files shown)


crates/kas-widgets/src/filler.rs
```diff
@@ -14,7 +14,7 @@ mod Filler {
     /// This widget has zero minimum size but can expand according to the given
     /// stretch priority.
     #[derive(Clone, Debug, Default)]
-    #[widget(type Data = ())]
+    #[widget]
     pub struct Filler {
```
dhardy, line 18: No `Data` def needed here now. Since `Data` cannot be used anywhere, the restriction to `()` is unambiguous.

crates/kas-widgets/src/float.rs
```diff
@@ -89,6 +89,7 @@ mod Float {
     #[widget]
     pub struct Float<C: Collection> {
         core: widget_core!(),
+        #[collection]
         widgets: C,
```
dhardy, line 93: New `#[collection]` attribute replaces two `Tile` fns and `impl Widget { ... }`.

PR #512, opened 2025-06-23
AccessKit integration

dhardy:
This PR adds basic [AccessKit](https://accesskit.dev/) integration.

WIP:

- [x] accesskit-winit adapter
- [x] framework for recursive update of initial tree
- [ ] tree updates
- [x] basic accesskit implementations for widgets
- [ ] handle AccessKit actions

Caveats:

- Text objects are handled in a naive manner (see also https://github.com/AccessKit/accesskit/issues/578)
- Many widgets do not set the correct properties for their Role, or do not set any properties at all; some roles may be incorrect. https://github.com/AccessKit/accesskit/issues/582 would help with this; user testing will also be important.
- Probably more ...

PR #519, opened 2025-07-14
Widget roles, labels

dhardy:
Let widgets report their role in a generic fashion not tied to AccessKit (see also #512).

Why this approach over #512?

- It doesn't tie us to AccessKit. I'm not convinced that AccessKit's current model is ideal; e.g. see https://github.com/AccessKit/accesskit/issues/591.
- Making this type of introspection a first-party citizen leads to re-evaluating some other design choices; for example `Mark` and `MarkButton` widgets now require an explicit label.
- This introspection has other potential uses, e.g. programmatic testing, debug diagnostics.

I'm not completely convinced by this approach yet, but think it should be enough to build a basic AccessKit implementation over.

Other changes:

- Fixed possible hash collision with allocated `Id` variants' `u64` representation (as far as is possible).
- Widgets `Mark`, `MarkButton` now require a label on construction. This is independent of `MarkStyle` since e.g. a down-chevron might be used to open a menu, minimize an application or decrement a value. ("Marks" are intended to be generic "glyphs" usable throughout a UI, not specific icons. Might need a rethink later.)


crates/kas-core/src/core/tile.rs
```diff
@@ -83,6 +83,27 @@ pub trait Tile: Layout {
...
+    /// Specify additional role properties for child `index`
+    ///
+    /// Role properties (for example [`RoleCxExt::set_label`]) may be specified
+    /// by a widget for itself in [`Self::role`] or by a parent in this method.
+    ///
+    /// The default implementation does nothing.
+    #[inline]
+    fn role_child_properties(&self, cx: &mut dyn RoleCx, index: usize) {
+        let _ = (cx, index);
+    }
```
dhardy, line 105: This is a slightly strange design, but allows determination of any widget's role + role properties without recursing the whole tree.

(For finding those children, we already have `child_indices` and `get_child`.)

crates/kas-widgets/src/menu/submenu.rs
```diff
@@ -54,7 +54,7 @@ mod SubMenu {
             SubMenu {
                 core: Default::default(),
                 label: AccessLabel::new(label).with_class(TextClass::MenuLabel),
-                mark: Mark::new(MarkStyle::Point(direction)),
+                mark: Mark::new(MarkStyle::Chevron(direction), "Open"),
```
dhardy, line 57: Example of a `Mark` with label.

crates/kas-widgets/src/scroll_label.rs
```diff
@@ -79,6 +79,16 @@ mod ScrollLabel {
...
     impl Tile for Self {
+        fn role(&self, _: &mut dyn RoleCx) -> Role<'_> {
+            // TODO: this is both a ScrollRegion and Text!
+            Role::Text {
+                text: self.text.as_str(),
+                editable: false,
+                edit_pos: self.selection.edit_pos(),
+                sel_pos: self.selection.sel_pos(),
+            }
+        }
```
dhardy, line 90: We can't use `Role::ScrollRegion` and `Role::Text` at the same time. We also can't construct a hierarchy of nodes that doesn't correspond to widgets here (`get_child` needs to return a `Tile` at least, if not a widget).

But I think it is acceptable:

1. The fact that `ScrollLabel` (and `ScrollText`, `EditField`) both scroll and have text is an odd design that limits code-reuse, so refactoring these seems reasonable.
2. We provide the API we need for widgets. If we want to allow introspection of complex text/SVG/canvas/... state later, we can add API appropriate for that (and with the appropriate methods; e.g. coordinate-to-text-position translation should be done through existing text methods, not by exporting a list of all glyph positions).

Issue #509, opened 2025-06-15
Tracker: a11y / accessibility tools


dhardy:
- [ ] Screen reader / external widget introspection support

This likely means using https://github.com/AccessKit/accesskit/

dhardy:
Known missing widget details from the current implementation:

- Text widgets do not report glyph positions to AccessKit (and thus do not support coordinate-to-text-index translation in accessibility tools)
- Text widgets do not report text selection
- Lists with selectable children should use role `List` and place their children in a `ListItem` node
- Tables should use role `Table`, have `RowHeader`s and `ColumnHeader`s and wrap children with `Cell`s; we don't have a real table widget yet though

DataTriny:
> Lists with selectable children should use role List and place their children in a ListItem node

If you're talking about AccessKit roles, the ones to use here are `Role::ListBox` and `Role::ListBoxOption`. You'll also want to call `Node::set_size_of_set` on the listbox node and `Node::set_position_in_set` on the items.

dhardy:
#521 adds AccessKit integration. Known issues:

- [x] #522
- [x] Some widgets should use `accesskit::Role::GenericContainer`
- [ ] [Text input content should use children with `Role::TextRun`](https://github.com/kas-gui/kas/pull/521#issuecomment-3094496650)
- [ ] #525 
- [x] Navigable widgets should advertise support for `accesskit::Action::Focus`
- [x] [Should all widgets support `Action::ScrollIntoView`](https://github.com/kas-gui/kas/pull/521#issuecomment-3094529361)?
- [x] Lists of selectable children should use `Role::ListBox` (see above comment)

dhardy:
@DataTriny is `Role::MenuItemCheckBox` substantially different from `CheckBox`? What expectations are there on `MenuItem` items?

dhardy:
> If you're talking about AccessKit roles, the ones to use here are `Role::ListBox` and `Role::ListBoxOption`. You'll also want to call `Node::set_size_of_set` on the listbox node and `Node::set_position_in_set` on the items.

[...]

PR #521, opened 2025-07-18
AccessKit integration (mark 2)

dhardy:
Replaces #512. In comparison:

- Widget code does not use AccessKit as an API; instead #519 added `fn Tile::role()`. This will doubtless need some additions, but lets widgets report necessary details without dependency on AccessKit (and with a nicer API in my opinion).
- Output to AccessKit should be largely the same as #512.
- Some more actions are supported vs #512.

This PR still needs a few tweaks (TEMP / TODO commits) but is testable.

@DataTriny I'd appreciate if you gave this a little test and some feedback. I have successfully tested with a screen reader. Any obvious issues? My notes:

- I didn't use `accesskit::Role::GenericContainer` at all here; just `Unknown` for layout containers. Is that acceptable?
- I notice some focus tracking issues in Kas when using access keys. Edit: cause is that the event handler tries to focus the `AccessLabel` which doesn't support focus. I'm not going to solve the issue in this PR.
- More known issues are noted in #509

I also am fairly convinced that AccessKit's model is not the best fit for Kas; see https://github.com/AccessKit/accesskit/issues/591. Perhaps a lower-level intermediate library or perhaps the best answer would just be to support platform-specific accessibility tools eventually.

### DIFF
--- a/crates/kas-core/src/core/role.rs
+++ b/crates/kas-core/src/core/role.rs
@@ -91,8 +91,9 @@ pub enum Role<'a> {
     ///
     /// ### Messages
     ///
-    /// [`kas::messages::SetValueString`] may be used to replace the entire
-    /// text.
+    /// [`kas::messages::SetValueText`] may be used to replace the entire
+    /// text. [`kas::messages::ReplaceSelectedText`] may be used to insert text
+    /// at `edit_pos`, replacing all text between `edit_pos` and `sel_pos`.
     TextInput {
         /// Text contents
         ///
--- a/crates/kas-core/src/core/role.rs
+++ b/crates/kas-core/src/core/role.rs
@@ -210,6 +210,120 @@ impl From<Id> for TextOrSource<'static> {
     }
 }
 
+#[cfg(feature = "accesskit")]
+impl<'a> Role<'a> {
+    /// Construct an AccessKit [`Role`] from self
+    pub(crate) fn as_accesskit_role(&self) -> accesskit::Role {
+        use accesskit::Role as R;
+
+        match self {
+            // TODO: do we want to automatically use role GenericContainer?
+            // Role::Unknown if has_children => R::GenericContainer,
+            Role::Unknown => R::Unknown,
+            Role::Label(_) | Role::AccessLabel(_, _) | Role::TextLabel { .. } => R::Label,
+            Role::Button => R::Button,
+            Role::CheckBox(_) => R::CheckBox,
+            Role::RadioButton(_) => R::RadioButton,
+            Role::Tab => R::Tab,
+            Role::Border => R::Unknown,
+            Role::ScrollRegion { .. } => R::ScrollView,
+            Role::ScrollBar { .. } => R::ScrollBar,
+            Role::Indicator => R::Unknown,
+            Role::Image => R::Image,
+            Role::Canvas => R::Canvas,
+            Role::TextInput {
+                multi_line: false, ..
+            } => R::TextInput,
+            Role::TextInput {
+                multi_line: true, ..
+            } => R::MultilineTextInput,
+            Role::Slider { .. } => R::Slider,
+            Role::SpinButton { .. } => R::SpinButton,
+            Role::ProgressBar(_) => R::ProgressIndicator,
+            Role::MenuBar => R::MenuBar,
+            Role::Menu { .. } => R::Menu,
+            Role::ComboBox { .. } => R::ComboBox,
+            Role::Window => R::Window,
+            Role::TitleBar => R::TitleBar,
+        }
+    }
+
+    /// Construct an AccessKit [`Node`] from self
+    ///
+    /// This will set node properties as provided by self, but not those provided by the parent.
+    pub(crate) fn as_accesskit_node(&self) -> accesskit::Node {
+        use crate::cast::Cast;
+
+        let mut node = accesskit::Node::new(self.as_accesskit_role());
+
+        match *self {
+            Role::Unknown | Role::Button | Role::Tab | Role::Border => (),
+            Role::Indicator | Role::Image | Role::Canvas => (),
+            Role::MenuBar | Role::Window | Role::TitleBar => (),
+            Role::Label(text) | Role::TextLabel { text, .. } | Role::TextInput { text, .. } => {
+                node.set_value(text)
+            }
+            Role::AccessLabel(text, ref key) => {
+                node.set_value(text);
+                if let Some(text) = key.to_text() {
+                    node.set_access_key(text);
+                }
+            }
+            Role::CheckBox(state) | Role::RadioButton(state) => node.set_toggled(state.into()),
+            Role::ScrollRegion { offset, max_offset } => {
+                node.set_scroll_x(offset.0.cast());
+                node.set_scroll_y(offset.1.cast());
+                node.set_scroll_x_min(0.0);
+                node.set_scroll_y_min(0.0);
+                node.set_scroll_x_max(max_offset.0.cast());
+                node.set_scroll_y_max(max_offset.1.cast());
+            }
+            Role::ScrollBar {
+                direction,
+                value,
+                max_value,
+            } => {
+                node.set_orientation(direction.into());
+                node.set_numeric_value(value.cast());
+                node.set_min_numeric_value(0.0);
+                node.set_max_numeric_value(max_value.cast());
+            }
+            Role::Slider {
+                min,
+                max,
+                step,
+                value,
+            }
+            | Role::SpinButton {
+                min,
+                max,
+                step,
+                value,
+            } => {
+                if min.is_finite() {
+                    node.set_min_numeric_value(min);
+                }
+                if max.is_finite() {
+                    node.set_max_numeric_value(max);
+                }
+                if step.is_finite() {
+                    node.set_numeric_value_step(step);
+                }
+                node.set_numeric_value(value);
+            }
+            Role::ProgressBar(value) => {
+                node.set_max_numeric_value(1.0);
+                node.set_numeric_value(value.cast());
+            }
+            Role::ComboBox { expanded, .. } | Role::Menu { expanded } => {
+                node.set_expanded(expanded);
+            }
+        }
+
+        node
+    }
+}
+
 /// Context through which additional role properties may be specified
 ///
 /// Unlike other widget method contexts, this is a trait; the caller provides an
--- a/crates/kas-core/src/core/role.rs
+++ b/crates/kas-core/src/core/role.rs
@@ -266,10 +266,11 @@ impl<'a> Role<'a> {
     /// Construct an AccessKit [`Node`] from self
     ///
     /// This will set node properties as provided by self, but not those provided by the parent.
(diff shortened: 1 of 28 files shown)


DataTriny:
Tested on Windows 11 with the NVDA screen reader. I don't exactly recall last time I gave kas a try but it roughly feels the same.

List of things I found are missing:
- Bounds: pointing at widgets with the mouse does not read their name, so widget bounds might not properly be set up.
- `Action::Click`: most widgets that can be interacted with should support it to invoke the "default action". This is particularly important on macOS where VoiceOver users usually will rely on VoiceOver commands to perform actions.
- Text input content and selection: I can see that the value of text inputs is properly set, but children with `Role::TextRun` are necessary for ATs to make sense of the content. Calling `Node::set_text_selection` on the text input node is required for ATs to report caret and selection updates (this also requires the existence of text runs).

> I didn't use accesskit::Role::GenericContainer at all here; just Unknown for layout containers. Is that acceptable?

No it is not: `Role::GenericContainer` is a "presentation" role and we filter it out of the tree that we send to the accessibility stacks. By using something else to represent panels which only purpose is for layout, you will polute the UI with nodes that users may thing are important but not labelled.

crates/kas-core/src/core/role.rs
```diff
@@ -158,6 +225,124 @@ impl From<Id> for TextOrSource<'static> {
...
+                node.set_value(text)
+            }
+            Role::AccessLabel(text, ref key) => {
+                node.set_value(text);
+                if let Some(text) = key.to_text() {
+                    node.set_access_key(text);
+                }
+            }
+            Role::CheckBox(state) | Role::RadioButton(state) => node.set_toggled(state.into()),
+            // node.set_toggled(match self.state {
```
DataTriny, line 288: The latest accesskit version contain you change to convert from a `bool` to a `Toggled`.

dhardy:
> Bounds

Fixed.

> `Action::Click`

This should work... can you give some specific examples? I wonder if the wrong target receives the message.

Note also that it might be useful to enable logging (`RUST_LOG=kas=debug`) and press F8 to get identifiers of widgets.

DataTriny:
Any chance of having the `accesskit` feature enabled by default?

dhardy:
> > I didn't use accesskit::Role::GenericContainer at all here; just Unknown for layout containers. Is that acceptable?
> 
> No it is not: `Role::GenericContainer` is a "presentation" role and we filter it out of the tree that we send to the accessibility stacks. By using something else to represent panels which only purpose is for layout, you will polute the UI with nodes that users may thing are important but not labelled.

I'm not quite sure of the best way to handle this. Some ideas:

1. Every `Unknown` node should be ignored since it specifies nothing useful about itself?
2. Report `Unknown` nodes with children as `Role::GenericContainer` since this is most likely what they are. (A lot of Kas widgets are macro-generated single-usage layout widgets.)
3. Add a specific `Role` in Kas. Macro-generated widgets would use this; others might still be reported as `Unknown`.

dhardy:
> Any chance of having the `accesskit` feature enabled by default?

Good question. I guess it should be. I don't want to commit to AccessKit, but since it's not part of the Kas API it's not much of a commitment. :+1:

DataTriny:
The bounds appear to be fixed, good job!

> This should work... can you give some specific examples? I wonder if the wrong target receives the message.

If I use the NVDA's keyboard shortcut to invoke a widget on the "Press me" button in the gallery example, it reports that there is no action to perform and I don't see the visual effect that I can observe if I click it with the mouse. Nothing relevant is found in the logs either.

DataTriny:
> Every Unknown node should be ignored since it specifies nothing useful about itself?

Not necessarily: these nodes could still have a label, support some actions and even use a role description to appear as a custom widget under rare circumstances.

> Report Unknown nodes with children as Role::GenericContainer since this is most likely what they are. (A lot of Kas widgets are macro-generated single-usage layout widgets.)

This should cover most cases I guess. Although you still risk having someone create a custom button with a child label and forgetting to set the proper node, you would now have a button that AT users wouldn't be able to reach.

dhardy:
> * Text input content and selection: I can see that the value of text inputs is properly set, but children with `Role::TextRun` are necessary for ATs to make sense of the content. Calling `Node::set_text_selection` on the text input node is required for ATs to report caret and selection updates (this also requires the existence of text runs).

Is this true only for `TextInput` roles or also for `Label`? In Kas I have both selectable and non-selectable variants of label.

What properties should be set on the `TextRun`: just bounds and label?

Does a `ReplaceSelectedText` action target the `TextRun` or the parent `TextInput`?

It turns out this is not so simple to integrate into Kas, so I may leave until later.

dhardy:
@DataTriny can you use https://github.com/kas-gui/kas/tree/push-uwyntlylqono (or just apply https://github.com/kas-gui/kas/commit/6d02785478e4bb7a603d475d6853960e4497663e), invoke the "Press me" button again, and report what gets printed to the console?

DataTriny:
I think I know what is wrong with actions: I didn't find any calls to `Node::add_action` in the code base. You have to tell upfront to which actions widgets can react to. This is important because some ATs will use this information to determine how to present a widget to the user. MacOS's VoiceOver uses this extensively for instance.

dhardy:
I don't really understand `Role::GenericContainer`. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/presentation_role talks about removing the semantics (such as heading style), but none of the layout widgets I would report using this would have any semantics anyway.

dhardy:
What do actions `Focus` and `Blur` refer to?

Is `SetSequentialFocusNavigationStartingPoint` important? I have methods to set focus (if supported) and set focus to the next/previous item, but not this exactly.

dhardy:
I'm not sure if I should advertise *every* node as supporting `ScrollIntoView`? I don't track what is scrollable while constructing the `Node` tree, but I can basically handle this on anything (without a `ScrollHint`; I don't have a good way to handle that). I can also scroll to ensure any point (`ScrollToPoint`) or rect is visible (assuming the scroll windows are sufficiently large).

But: just reporting everything as supporting those actions seems wrong?

DataTriny:
> Is this true only for TextInput roles or also for Label? In Kas I have both selectable and non-selectable variants of label.

Only text input roles for now. This will change when we will support web documents but we are far from getting there.

> What properties should be set on the TextRun: just bounds and label?

`Node::bounds`, `Node::character_lengths`, `Node::character_positions`, `Node::character_widths`, `Node::text_direction`, `Node::value`, `Node::word_lengths` (optional).

> Does a ReplaceSelectedText action target the TextRun or the parent TextInput?

The parent text input.

DataTriny:
> What do actions Focus and Blur refer to?

It basically mean that they can gain or lose focus. We don't yet implement the blur action, but if a widget is part of the focus chain, it must have `Action::Focus` and you should give it focus if you receive this action request from AccessKit. This is not used so much on Windows and Linux where screen reader users typically use the keyboard navigation mechanism provided by the apps, but this is how macOS users navigate UIs. Users relying on voice control would also almost exclusively use this.

> Is SetSequentialFocusNavigationStartingPoint important?

We don't yet implement it. Please ignore it for now.

> I'm not sure if I should advertise every node as supporting ScrollIntoView?

We recently added `Node::add_child_action` which can be applied to nodes with `Role::ScrollView`. This is a more convenient way for you to say that descendents of this node all support `Action::ScrollIntoView`.

dhardy:
Advertising `Action::Focus` I can do, but it requires changes to widget traits, so I'd rather use a new PR.

`Node::add_child_action` notes that this adds the action to *direct* children. Scrolling-into-view is handled recursively in Kas, so really it applies to *everything*, though some additional code would be needed to work out where it's actually useful.

dhardy:
Oh, clicking the "Press me" button should now work. Also `Expand`-ing the ComboBox menu and top menus. Except the menus won't get keyboard focus normally, so I think the only ways to open the menus are to click or to use access keys.

DataTriny:
> Oh, clicking the "Press me" button should now work. Also Expand-ing the ComboBox menu and top menus. Except the menus won't get keyboard focus normally, so I think the only ways to open the menus are to click or to use access keys.

Yes, buttons can now be activated. I am not testing the menus because we haven't implemented them yet, as well as access keys. I suppose you could easily implement `Action::Decrement` and `Action::Increment` for sliders and spin boxes.

dhardy:
> I suppose you could easily implement `Action::Decrement` and `Action::Increment` for sliders and spin boxes.

I forgot that one. Done.

Then I'll merge with known issues. Thanks for the help.

DataTriny:
I can see in the gallery example that there is a tab list/side bar. Things that are missing for this pattern: the container must be of role `Role::TabList` and have `Node::size_of_set`. Children are missing `Node::position_in_set` and `Node::selected`. The widget that each tab is controlling must be `Role::TabPanel`. To do things properly `Node::controls` for the tab should contain the `NodeId` of the panel it is controlling.

dhardy:
@DataTriny that's tricky since I don't have a widget corresponding to that `Role::TabList`. All `Tab` widgets are children of the `TabStack` widget. I could maybe change that.

I also don't have anything corresponding to a `TabPanel`. Why is this needed?

DataTriny:
> I also don't have anything corresponding to a TabPanel. Why is this needed?

Semantic: to know what is linked to the tab and what is not. If you navigate using the keyboard and if there are widgets after the tab panel, you wouldn't know where it ends.

dhardy:
> > I also don't have anything corresponding to a TabPanel. Why is this needed?
> 
> Semantic: to know what is linked to the tab and what is not. If you navigate using the keyboard and if there are widgets after the tab panel, you wouldn't know where it ends.

So is `Role::TabPanel` any different from `ListItem` or `Pane` or `Cell`? Some of these might support selection, but selection is an action... (I'm not sure if there is a corresponding AccessKit action). Otherwise, I don't think there's any reason for `TabPanel` to be a separate role from `Pane` or `Cell` or `LayoutTableCell`?

DataTriny:
> Some of these might support selection, but selection is an action...

A tab panel does not support selection, it is a container. Only its related tab support selection. Furthermore, selection is a state, represented in AccessKit by `Node::selection`: if the widget is currently in a selected state you'd set this to `true`, if it is not, you set it to `false`, and if the widget does not support the selection state then you should not call `Node::set_selected` at all. A widget's selection state can be updated as a result of `Action::Click`.

You can view `TabPanel` as a specialized `Pane`. Again, it is here to express the fact that it is linked to a `Tab` in a tab pattern. Just like a `Cell` only make sense in the context of a table. As a UI toolkit developer, your role is to give app authors the necessary tools to convey the meaning of their UI, not just give them the tools to make something beautiful. This is not something specific to AccessKit by the way, this rich data schema is here so that AT developers can build accessible experiences for their users to be more productive.

DataTriny:
To recap: the "tab pattern" is comprized of a `TabList` and one or more `TabPanel`s. Typically only one `TabPanel` will be shown at a time. The `TabList` contain one ore more `Tab`s, each corresponding to a `TabPanel`. The `Tab`s support the selection state.

PR #526, opened 2025-07-22
Add Role::None, Grip, Splitter, OptionList{,Item}, Grid{,Cell}; nightly-diagnostics-pedantic

dhardy:
New roles:

- `Grip` makes the rect of movable handles (e.g. of a slider) accessible. AccessKit does not use this.
- `None` makes explicit that the widget has no semantic role and should be ignored. This is now the default for widgets with macro-defined layout and children.
- `Splitter` for the widget with the same name... but this achieves nothing, so I'm not sure it should be included now. Should this do something?
- `OptionList`, `OptionListItem` for `ListView` which may have selectable children. This could (but does not) expose the selection mode. Children use `Node::set_selected` only if true or if the selection mode is multiple, which is a compromise.
- `Grid`, `GridCell` report the grid size, cell locations and selection status for `GridView`.

Removed:

- Widget `NavFrame`, view driver `NavView`.

New feature gate `nightly-diagnostics-pedantic`: this prints warnings about missing `role` definitions, with an exception for those defaulting to `None` (see above). This lint is denied in libs but not in examples:
```
$ cargo check --features nightly-diagnostics-pedantic --examples
warning: [pedantic] `fn Tile::role` is not defined
  --> examples/proxy.rs:53:5
   |
53 |     #[widget]
   |     ^^^^^^^^^

warning: [pedantic] `fn Tile::role` is not defined
  --> examples/clock.rs:36:5
   |
36 |     #[widget]
   |     ^^^^^^^^^
```

My testing:

- `OptionList` tells me the list has length 34 items. Console debugging tells me the real length: 24307 items. Why? Maybe it's just Orca.
- `OptionListItem` tells me the selection status, ~~but nothing else. AccessKit should traverse the tree to read the contents? I mean, this isn't C or GTK; this is Rust which supports generics; the `ListView` doesn't know what the contents are. I could call `set_labelled_by` but this would fail for `examples/data-list-view.rs` where list items are a complex widget, not a simple label.~~ See below.
- `Grid`, `GridCell` ~~tell me nothing (the screen reader says "blank"). As above, AccessKit needs to recurse the node tree to read contents instead of expecting the cell to have a label / value.~~ But also, should it not at least tell me that this is a grid with 12 columns and rows? (Try `examples/times-tables.rs`.)

CC @DataTriny


dhardy:
> * `OptionListItem` tells me the selection status, but nothing else. AccessKit should traverse the tree to read the contents? I mean, this isn't C or GTK; this is Rust which supports generics; the `ListView` doesn't know what the contents are. I could call `set_labelled_by` but this would fail for `examples/data-list-view.rs` where list items are a complex widget, not a simple label.

The screen reader actually does reasonably well in `examples/data-list-view.rs` where the list entry contains an `EditField` and a `RadioButton` which can be focussed. So the problem is more how to handle the case where the list item / grid cell is just a label.

Unfortunately Rust doesn't yet support specialization, so we can't `set_labelled_by` only in that case. (And even if we could, it would be brittle; a user might make their own label widget.)

Edit: just setting `labelled_by` in all cases seems to work in my testing. This is a hack though and could fail, e.g. if the contents are a layout (presentation) widget around a label and icon.

PR #528, opened 2025-07-25
kas-text update; use child Role::TextRun for input fields

dhardy:
This updates kas-text corresponding to https://github.com/kas-gui/kas-text/pull/98.

Adds `Role::AccesskitVerbatim` and uses it to report the text in an `EditField` as [specified here](https://github.com/kas-gui/kas/pull/521#issuecomment-3094529514), save that character and word data is not provided (requires more extensive `kas-text` changes I don't wish to work on since in my opinion AccessKit should query existing text objects for this kind of detail instead of using its own algorithms over character-level reconstructions).

This is a hack and hopefully will be replaced eventually.

In my testing, the screen reader can now read `SpinBox` values but not `EditBox` values; I don't know why but the only difference visible to AccessKit should be what the parent node is (`Role::SpinButton` in the former case, `Role::ScrollView` in the latter case). CC @DataTriny


DataTriny:
At least for now and on Windows, all text inputs appear as empty to screen readers even if they are not, and moving the caret report nothing. Perhaps more interestingly at this stage, I've noticed a delay when tabbing into a text input.

dhardy:
I don't understand what's going on there. There shouldn't be a delay; if anything AccessKit info is updated far too often in the current code. All text inputs should get their contents set via a `TextRun` child too.

dhardy:
I removed `Role::AccesskitVerbatim` from this PR since it failed to make `EditField` input accessible and is not my preferred approach. The commits are here: https://github.com/kas-gui/kas/compare/b28d7477..accesskit-verbatim

PR #535, opened 2025-08-11
Make `fn DataClerk::len` optional for unknown-length scrollable views

dhardy:
Changes:

- Replace `fn Scrollable::scroll_axes()` with `content_size()` (simpler to specify and implement). The method could *almost* be removed entirely, but there is no correct alternative for auto-detecting requirement for scroll bars in `set_rect`.
- Make `fn DataClerk::len()` return an `Option`; add `fn DataClerk::min_len(expected: Index) -> Index`. This allows usage of `ListView` & `GridView` without a known data length (or without advertising the length).
- Add a scroll bar to the gallery list. Make the row limit optional in `examples/data-list-view.rs`.

Fixes:

- `child_to_data` caused incorrect Tab-navigation in `GridView`
- The initial window sizing on Wayland (and potentially other platforms) is improved. Unfortunately it's still wrong since we are forced to use logical pixels and are told the wrong scale factor when checking available monitors (in my case, this shows 200% but I use 130%) but from what I understand this is the best possible under Wayland.

### DIFF
--- a/crates/kas-view/src/grid_view.rs
+++ b/crates/kas-view/src/grid_view.rs
@@ -402,11 +402,13 @@ mod GridView {
             let end_row = u32::conv(u64::conv(view_end_offset.1) / u64::conv(skip.1) + 1)
                 .min(min_data_len.row);
 
-            if !(force
-                || first_col < self.first_data.col
-                || first_row < self.first_data.row
-                || end_col > self.first_data.col + self.cur_len.col
-                || end_row > self.first_data.row + self.cur_len.row)
+            if !force
+                && first_col >= self.first_data.col
+                && first_row >= self.first_data.row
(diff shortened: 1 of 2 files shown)


examples/data-list-view.rs
```diff
@@ -152,8 +159,13 @@ impl DataClerk<usize> for Clerk {
...
-    fn len(&self, data: &Self::Data) -> usize {
-        data.len
+    fn len(&self, data: &Self::Data) -> Option<usize> {
+        data.row_limit.then_some(data.len)
+    }
+
+    fn min_len(&self, _: &Self::Data, expected: usize) -> usize {
+        // This method is only called when len returns None, i.e. when we don't use data.len
+        expected
     }
```
dhardy, line 169: len / min_len example

PR #542, opened 2025-08-19
Access keys: register on draw

### LINKED ISSUE #536: Access keys should be restricted to visible UI elements
Currently, access keys are registered when widgets are configured and assigned to "layers". These layers are only really used for pop-up menus, not e.g. tab stacks.

Access keys should not be usable on inactive tab stack pages at least.

dhardy:
Previously access keys were registered as widgets were configured, then bound to layers per pop-up. Giving stack pages their own layer helped but did not solve issues.

Determining visible widgets through `Tile::child_indices` and registering access keys from that was also considered, but found insufficient.

It turns out that building a hash-set of access-key registrations on each draw is not particularly expensive. In fact, my attempt to benchmark this showed *improved* draw times (slightly suspicious):
```
1. cargo run --example gallery --features nightly-diagnostics,stable --release
2. Tap Alt key repeatedly
3. Copy last 100 lines into 'out'
4. cat out | cut -b 62-64 | paste -sd+ | bc
Before: 59095, 58448
After: 57224, 57743
```
Results are the sum of 100 draws in μs, as reported by `do_draw` under `kas_perf`.

Closes #536.

### DIFF
--- a/crates/kas-core/src/config/config.rs
+++ b/crates/kas-core/src/config/config.rs
@@ -92,6 +92,7 @@ pub struct WindowConfig {
     pub(super) kinetic_decay_sub: f32,
     pub(super) scroll_dist: f32,
     pub(super) pan_dist_thresh: f32,
+    pub(crate) alt_bypass: bool,
     /// Whether navigation focus is enabled for this application window
     pub(crate) nav_focus: bool,
 }
@@ -107,6 +108,7 @@ impl WindowConfig {
             kinetic_decay_sub: f32::NAN,
             scroll_dist: f32::NAN,
             pan_dist_thresh: f32::NAN,
+            alt_bypass: false,
             nav_focus: true,
         }
     }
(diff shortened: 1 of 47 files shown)


dhardy:
Changelog:

- Hidden enum `kas::NavAdvance` moved to `kas::event::NavAdvance`
- Removed `EventState` fns `show_access_labels, new_access_layer, enable_alt_bypass, add_access_key, disable_nav_focus`
- Added `fn EventState::request_nav_focus`, replacing `set_nav_focus` (which is now crate-private)
- Renamed `fn Event::on_activate` → `on_click`
- New `DrawCx` fns `access_key, text_with_effects`
- Removed `align` parameter from `fn Popup::new` and replaced `fn Popup::set_alignment` with `fn align` (inline)
- Added `Window` fns `with_alt_bypass, without_nav_focus` (inline)
- Added `fn AccessLabel::set_target`
- Removed `fn AdaptConfigCx::enable_alt_bypass`
- Removed `TextClass` parameter from `trait ThemeDraw`'s text methods

Alt-bypass mode can now only be enabled at the window level. A more fine-tuned approach could be supported (e.g. by attaching a `bool` to each access key registration) but is not currently motivated.

PR #563, opened 2025-09-26
Move fn Tile::probe to new trait LayoutFnProbe

dhardy:
`fn probe` did not really belong in the `Tile` trait, especially with the `Self: Sized` bound. It is now moved to a new trait.

Method `fn probe` is now implemented in `Layout` but moved to an implementation of `LayoutFnProbe`, similar to how `type Data` may be moved from `Events` to `Widget`.


dhardy:
Documentation of "foreign impl items" is a little worse than for native items; I added this to the trait docs (`Events` here; similarly in `Layout`):
image

dhardy:
Significantly, this now supports non-rectangular hit testing (e.g. `Button` could reject clicks on the corners of its rect if we added code to get the corner radius from the theme; this is not implemented here).

PR #572, opened 2025-10-04
Mut text access, message sending doc

dhardy:
See commits

### DIFF
--- a/crates/kas-widgets/src/text.rs
+++ b/crates/kas-widgets/src/text.rs
@@ -28,7 +28,7 @@ mod Text {
     pub struct Text<A, T: Default + FormattableText + 'static> {
         core: widget_core!(),
         text: theme::Text<T>,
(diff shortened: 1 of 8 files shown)


crates/kas-widgets/src/text.rs
```diff
@@ -39,19 +39,68 @@ mod Text {
...
 
     impl Self {
-        /// Construct with a data binding
-        #[inline]
-        pub fn new(text_fn: impl Fn(&ConfigCx, &A) -> T + 'static) -> Self {
+        /// Construct with a generator function
+        ///
+        /// `gen_text` is called on each widget update to generate text from
+        /// input data.
+        pub fn new_gen(gen_text: impl Fn(&ConfigCx, &A) -> T + 'static) -> Self {
```
dhardy, line 77: Breaking change: `Text::new` is renamed `Text::new_gen`; additional constructors added: `new_str` and `new_update`

PR #594, opened 2025-11-11
Add trait Viewport

dhardy:
#593 highlights the issues with `trait Scrollable` but does not offer a compelling alternative.

#593 also misses that `ScrollText` is a separate widget from `SelectableText` for a reason: each must have a different `Role`. (This also applies to `ListView` but not `ScrollBarsRegion`.)

This PR offers an alternative:
```rust
pub trait Viewport: Layout {
    /// Get content size
    fn content_size(&self) -> Size;

    /// Update the scroll offset
    fn update_offset(&mut self, viewport: Rect, offset: Offset) {
        let _ = (viewport, offset);
    }

    /// Draw with a scroll offset
    fn draw_with_offset(&self, draw: DrawCx, viewport: Rect, offset: Offset);
}
```
Like `ScrollText`, the scroll offset is managed by the parent widget, not the content. I *think* this approach will extend reasonably well to `ListView` (except for the ridiculous `fn _nav_next` method).

### DIFF
--- a/crates/kas-widgets/src/edit/edit_field.rs
+++ b/crates/kas-widgets/src/edit/edit_field.rs
@@ -135,12 +135,11 @@ mod EditField {
     impl Viewport for Self {
         #[inline]
         fn content_size(&self) -> Size {
-            let mut size = self.rect().size;
             if let Ok((tl, br)) = self.text.bounding_box() {
-                size.1 = size.1.max((br.1 - tl.1).cast_ceil());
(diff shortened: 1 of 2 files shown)


dhardy:
~~rust-analyzer can be confused by this, reporting an error (false positive) when an explicit `Layout` impl without `fn draw` is used. And yet this is not new; for example the `Align` widget already impls `Layout` without `fn draw`.~~ No longer appears to be the case (for no obvious reason).

PR #595, opened 2025-11-11
Use Viewport in ScrollRegion

dhardy:
The renames `ScrollRegion` → `ClipRegion` and merges `ScrollBars` and `ScrollBarRegion` into a new widget named `ScrollRegion`.

The old `ScrollBars` remains for compatibility with `ListView` / `GridView` (until a future PR).

`ScrollRegion` is generic but constructed as a `ScrollRegion<ClipRegion<Inner>>` through the default constructor.


dhardy:
CI failure is only documentation on the soon-to-be-removed `ScrollBars` widget.

PR #603, opened 2025-11-22
Update IME code

dhardy:
New features:

- Expose surrounding text to IME
- Underline IME pre-edit (hacky impl)

Lots of small changes/fixes to IME and action handling (also affects selection action interaction with keyboard input). Has some testing with IBus on KDE/Wayland.

### DIFF
--- a/crates/kas-widgets/src/edit/mod.rs
+++ b/crates/kas-widgets/src/edit/mod.rs
@@ -34,10 +34,15 @@ enum EditAction {
 /// Used to track ongoing incompatible actions
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 enum CurrentAction {
+    /// No current action
     #[default]
     None,
+    /// IME is enabled but no input has yet been given. This is special in that
+    /// a selection may exist (which would get replaced by the pre-edit text).
     ImeStart,
-    /// We have a pre-edit text
+    /// We have some pre-edit text within the given range (if non-empty).
+    ///
+    /// This text should be deleted if IME is cancelled.
     ImePreedit {
         /// Range of the pre-edit text
         edit_range: Range<u32>,


crates/kas-core/src/event/cx/key.rs
```diff
@@ -201,42 +192,126 @@ impl EventState {
...
+    /// If `target` does not have selection focus this request will be ignored.
+    /// If `target` already has IME focus, it will be replaced (using the new
+    /// parameters provided).
+    pub fn request_ime_focus(
+        &mut self,
+        target: Id,
+        hint: ImeHint,
+        purpose: ImePurpose,
+        mut surrounding_text: Option<ImeSurroundingText>,
+    ) {
```
dhardy: All `IME` requests except `set_ime_cursor_area` are now on `EventCx` and require pre-existing selection focus. (All requests require the `target` be specified and ignore the request if this target does not have selection focus.)

crates/kas-core/src/event/event.rs
```diff
@@ -61,22 +118,18 @@ pub enum Event<'a> {
...
-    /// This is only received after
-    /// [requesting key focus](EventState::request_key_focus) with some `ime`
-    /// purpose.
-    ImeCommit(&'a str),
+    /// On [`Ime::Enabled`],
+    /// the widget should call [`EventState::set_ime_cursor_area`] immediately
+    /// and each time the area changes (relative to the widget's coordinate
+    /// space), until [`Ime::Disabled`] is received. Failure to do so will
+    /// result in the widget's entire `rect` being used as the IME cursor area.
+    Ime(Ime<'a>),
```
dhardy: All IME events are aggregated under `Event::Ime`; previously they weren't.

Motivation: a widget using IME should handle all IME events; other widgets can ignore them all.

crates/kas-widgets/src/edit/edit_field.rs
```diff
@@ -461,14 +538,17 @@ mod EditField {
...
-            }
             self.set_error_state(cx, false);
+
+            // NOTE: we would call self.cancel_selection_and_restart_ime() if we had an EventCx
+            if self.current == CurrentAction::Selection {
+                self.input_handler.stop_selecting();
+                self.current = CurrentAction::None;
+            } else if self.current.is_ime() {
+                cx.request_timer(self.id(), TIMER_RESTART_IME, Default::default());
+            }
```
dhardy: Key/mouse input calls `self.cancel_selection_and_restart_ime()` to handle on-going actions. `fn set_string` can't do this without an `EventCx`.

Instead we use a timer as a callback to restart IME in case `set_string` is called. Hacky but probably fine.

PR #615, opened 2025-12-15
`Tile::core{,_mut}` methods, `Flow` widget

dhardy:
Add methods `Tile::core` and `core_mut` where `Self: Sized`. These enable `derive_widget` to access the inner widget's core.

Revise `WidgetStatus` code (merge into `WidgetCore` trait). Add some missing status checks & updates for derived widgets.

Fix `AxisInfo::other` parameter (width estimate) provided by the `Window` widget. (These width estimates are not currently tested against assigned `Rect` widths but should be in debug builds. `FlowSolver` / `FlowSetter` have been designed to assume that this is correct, but avoid crashing in case they aren't.)

Adds a `Flow` widget. It has some limitations but supports horizontal rows:
image

Unfinished (excluded from this PR):
- proper padding and alignment of flow items
- support vertical (column-based) flows

### DIFF
--- a/crates/kas-core/Cargo.toml
+++ b/crates/kas-core/Cargo.toml
@@ -148,6 +148,8 @@ match_like_matches_macro = "allow"
 needless_range_loop = "allow"
 too_many_arguments = "allow"
 neg_cmp_op_on_partial_ord = "allow"
+collapsible_if = "allow"
+collapsible_else_if = "allow"
 
 [lints.rust]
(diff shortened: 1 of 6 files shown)


dhardy:
Change: I renamed `Flow::right_down` and `left_down` to `right` and `left` respectively, since it is natural to assume that lines wrap downwards. (There is still room for adding `down_left`, `down_right`, `up_right`, `right_up` etc.; though I doubt we'd ever want any of the `up` variants.)

PR #620, opened 2025-12-31
Subpixel text rendering; `RasterConfig` tweaks

dhardy:
Implements sub-pixel font rendering (currently limited to kas-soft only due to a wgpu validation issue).
image
image

`RasterConfig` changes:

- Remove `scale_steps`. This allowed the font size (dpem) to be fractional. I don't believe this is useful so have disabled it completely (but if necessary, re-enabling with a fixed precision is easy).
- Disable vertical-axis sub-pixel positioning. Since the line position is already clamped to the nearest pixel, this isn't really useful.
- Rename `subpixel_steps` → `subpixel_x_steps`
- Add `subpixel_mode` as an enum

Add this (currently only supporting one sub-pixel arrangement):
```rust
pub enum SubpixelMode {
    /// No sub-pixel rendering
    #[default]
    None,
    /// Horizontal RGB sub-pixels
    HorizontalRGB,
}
```

Sub-pixel filtering is implemented in kas-soft using [FreeType2's default LCD filter](https://freetype.org/freetype2/docs/reference/ft2-lcd_rendering.html): "beveled, normalized, and color-balanced five-tap filter with weights of [0x08 0x4D 0x56 0x4D 0x08] in 1/256 units". Since kas-soft knows exactly how the rendering works and the sub-pixel arrangement is fixed to `HorizontalRGB` it seems unnecessary to make the filter configurable.

kas-wgpu currently has most of the sub-pixel rendering implementation without filtering, but the whole thing is disabled because of an unresolved validation error.


dhardy:
I can confirm that sub-pixel rendering works with kas-wgpu using a small hack to bypass the validation issue. No filtering is implemented in this case so some coloured fringes may be visible:
image

Worth linking the resource that helped me work out how to use dual-source blending for sub-pixel rendering: https://arkanis.de/weblog/2023-08-14-simple-good-quality-subpixel-text-rendering-in-opengl-with-stb-truetype-and-dual-source-blending/#gl45-subpixel-text-rendering-gamma-1.43

PR #627, opened 2026-01-03
Add inner: &mut W parameter to Adapt closures and revise `DrawShared::image_*` fns

dhardy:
Draft because: I'm undecided whether this is a good design; the purpose of `Adapt` is to add user-controllable state.

### DIFF
--- a/crates/kas-core/src/draw/draw_shared.rs
+++ b/crates/kas-core/src/draw/draw_shared.rs
@@ -93,7 +93,7 @@ pub trait DrawShared {
     /// Allocate an image
     ///
     /// Use [`SharedState::image_upload`] to set contents of the new image.
-    fn image_alloc(&mut self, size: (u32, u32)) -> Result<ImageHandle, AllocError>;
+    fn image_alloc(&mut self, size: Size) -> Result<ImageHandle, AllocError>;
 
     /// Upload an image to the GPU
     ///
@@ -117,7 +117,7 @@ pub trait DrawShared {
 
 impl<DS: DrawSharedImpl> DrawShared for SharedState<DS> {
     #[inline]
-    fn image_alloc(&mut self, size: (u32, u32)) -> Result<ImageHandle, AllocError> {
+    fn image_alloc(&mut self, size: Size) -> Result<ImageHandle, AllocError> {
         self.draw
             .image_alloc(size)
             .map(|id| ImageHandle(id, Rc::new(())))
@@ -156,7 +156,7 @@ pub trait DrawSharedImpl: Any {
     /// Allocate an image
     ///
     /// Use [`DrawSharedImpl::image_upload`] to set contents of the new image.
-    fn image_alloc(&mut self, size: (u32, u32)) -> Result<ImageId, AllocError>;
+    fn image_alloc(&mut self, size: Size) -> Result<ImageId, AllocError>;
 
     /// Upload an image to the GPU
     ///
--- a/crates/kas-core/src/draw/draw_shared.rs
+++ b/crates/kas-core/src/draw/draw_shared.rs
@@ -7,6 +7,7 @@
 
 use super::color::Rgba;
 use super::{DrawImpl, PassId};
+use crate::ActionRedraw;
 use crate::cast::Cast;
 use crate::config::RasterConfig;
 use crate::geom::{Quad, Size, Vec2};
@@ -62,6 +63,17 @@ pub enum ImageFormat {
 #[error("failed to allocate: size too large or zero-sized")]
 pub struct AllocError;
 
+/// Upload failed
+#[derive(Error, Debug)]
+pub enum UploadError {
+    /// No allocation found for the [`ImageId`] used
+    #[error("image_upload: allocation not found")]
+    Missing,
+    /// Size of the uploaded image is wrong
+    #[error("image_upload: size does not match the allocation")]
+    Size,
+}
+
 /// Shared draw state
 ///
 /// A single [`SharedState`] instance is shared by all windows and draw contexts.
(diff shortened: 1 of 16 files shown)


dhardy:
I think `Sprite` could now be used something like this:
```rust
use kas::cast::Conv;
use kas::draw::ImageFormat;
use kas::event::TimerHandle;
use kas::geom::Size;
use kas::layout::LogicalSize;
use kas::Widget;
use kas_image::Sprite;
use kas_widgets::AdaptWidget;
use std::time::Duration;

const TIMER: TimerHandle = TimerHandle::new(0, true);

#[derive(Debug, Default)]
struct Buffer {
    size: Size,
    buffer: Vec<u8>,
}

async fn draw(mut buf: Buffer) -> Buffer {
    let len = usize::conv(buf.size.0) * usize::conv(buf.size.1);
    buf.buffer.resize(len, 0);
    // TODO: draw to buffer
    buf
}

enum State {
    Ready(Buffer),
    Drawing,
}

/// Make a sprite which draws itself in a worker thread on update
fn make_sprite(size: LogicalSize) -> impl Widget<Data = ()> {
    Sprite::new()
        .with_logical_size(size)
        .with_state(State::Ready(Buffer::default))
        .on_update(|cx, sprite, state, _| {
            if matches!(state, State::Ready(_)) {
                // We only have a ConfigCx here; use a timer to get an EventCx:
                cx.request_timer(TIMER, Duration::ZERO)
            }
        })
        .on_timer(TIMER, |cx, sprite, state, _| {
            if let State::Ready(buf) = std::mem::replace(state, State::Drawing) {
                if let Some(size) = sprite.allocate(cx) {
                    buf.size = size;
                    let id = cx.id();
                    cx.send_spawn(id, || draw(buf))
                }
            }
        })
        .on_message(|cx, sprite, buf| {
            if let Some(handle) = sprite.handle() {
                let draw = cx.draw_shared();
                if let Some(alloc_size) = draw.image_size(handle) && alloc_size == buf.size {
                    cx.draw_shared().image_upload(handle, &buf.buffer, ImageFormat::Rgba8);
                }
            }
        })
}
```
(I wanted to add this as example text, but cannot because `kas-image` does not depend on `kas-widgets`, while `kas_widgets::Adapt` is required here.)

I may merge this (or similar). @Kwarrtz can you test, or maybe update `examples/proxy` using an iterated function system?

Kwarrtz:
I think iterated function systems might be needlessly complex as an example, but I can try to come up with something to demonstrate using `Sprite`. Maybe just drawing a gradient with changeable colors?

I'm a bit confused about the purpose of `allocate` here though. It re-allocates if the size of the allocated buffer and the `Sprite`'s rect no longer match, but other than a call to `set_logical_size`, what could cause that to happen?

dhardy:
Kas widgets don't choose their own size; they "suggest" size requirements; as such the size may be changed e.g. by the user resizing the window (depending on whatever layout is used; usually fixed-size widgets do get their chosen size). Since this is "logical size", the size could also change if the scale factor changes, e.g. by dragging the window to another screen.

The state machine is to only draw on demand (i.e. on resize), and if not already drawing. In your case you may still need something like this to respond to resizes. Also, you might want to use a double buffer (one to draw into, another to read from) to avoid tearing and making `Mutex` unhappy (assuming you use `Arc<Mutex>` to share the buffers and not just pass them to/from the rendering thread as here).

dhardy:
I'm happy to merge this, though not completely happy about `allocate`; it's a bit awkward to use. Hence why having an example would be nice.

Kwarrtz:
Ah, I see, I'd actually somehow missed that `Sprite` supports non-fixed sizes. If one wanted to reallocated and redraw on resize to stay pixel perfect (which seems to me like the main use-case for `allocate`), am I correct to think that that would need to happen during a `set_rect` call? 

Also, if you're not happy how `allocate` works, maybe it could just be removed? If the redraw computation isn't instantaneous and is being forked off, like in your example, then it seems like you might want to defer calling `Sprite::set` until the new buffer is ready and so handle the re-allocation manually anyway. (Again, unless I'm misunderstanding the use case you had in mind for it.)

dhardy:
> If one wanted to reallocated and redraw on resize to stay pixel perfect (which seems to me like the main use-case for `allocate`), am I correct to think that that would need to happen during a `set_rect` call?

Yes. So `on_update` would need to be replaced with `on_set_rect` which doesn't exist yet. Good catch.

> Also, if you're not happy how `allocate` works, maybe it could just be removed?

The allocation needs to happen somewhere. But yes, `Sprite::allocate` is purely a convenience function which could be replaced by `DrawShared::image_alloc` and `Sprite::set`.

Personally I would choose to use a custom widget at this point, but it may be useful to provide enough tools to let users implement something like this without a custom widget.

dhardy:
Without `Sprite::allocate`, the above needs adapting to something like this (or a simpler version if you don't care about 
```rust
        .on_timer(TIMER, |cx, sprite, state, _| {
            if let State::Ready(buf) = std::mem::replace(state, State::Drawing) {
                let draw = cx.draw_shared();
                let size = sprite.rect().size;
                if let Some(handle) = sprite.handle() && draw.image_size(handle) == Some(size) {
                    return;
                }
                
                if let Ok(handle) = draw.image_alloc(size.cast()) && sprite.set(cx, handle) {
                    buf.size = size;
                    let id = cx.id();
                    cx.send_spawn(id, || draw(buf));
                }
            }
        })
```
You could simplify a bit, but should certainly not re-allocate each time this is called since there's no way to ensure `on_timer` is called only once.

Kwarrtz:
Yes, my thought was exactly that if a custom widget was needed for the `set_rect` behavior anyway, at that point calling `image_alloc` and `set` manually hardly seems prohibitively verbose. But if an `on_set_rect` adapter was added too, then maybe keeping the convenience function makes more sense. I'm just not sure how many applications there would be for that beyond this.

In either case, I'll try to put together a concrete example tomorrow.

dhardy:
This got several updates motivated by #631:

- Remove `fn Sprite::allocate`
- Add `fn Sprite::image_size`
- Let `image_alloc` use type `Size`
- Require passing `size` into `image_upload`
- Let `image_upload` return a `Result<ActionRedraw, UploadError>` to both notify of failures and nudge user to queue a redraw

PR #630, opened 2026-01-05
Support `row!` and `column!` within `grid!`, replacing `aligned_row!` and `aligned_column!`

dhardy:
The primary motivation is that this allows much nicer layout specification for `EventConfig`:
```rust
    #[layout(grid! {
        row!["Hover delay:", self.menu_delay],
        row!["Menu delay:", self.menu_delay],
        row!["Touch-selection delay:", self.touch_select_delay],
        row!["Kinetic scrolling timeout:", self.kinetic_timeout],
        row!["Kinetic decay (relative):", self.kinetic_decay_mul],
        row!["Kinetic decay (absolute):", self.kinetic_decay_sub],
        row!["Kinetic decay when grabbed:", self.kinetic_grab_sub],
        row!["Scroll wheel distance:", self.scroll_dist_em],
        row!["Pan distance threshold:", self.pan_dist_thresh],
        row!["Double-click distance threshold:", self.double_click_dist_thresh],
        row!["Mouse pan:", self.mouse_pan],
        row!["Mouse text pan:", self.mouse_text_pan],
        row![_, self.mouse_wheel_actions],
        row![_, self.mouse_nav_focus],
        row![_, self.touch_nav_focus],
        row![
            "Restore default values:",
            Button::label_msg("&Reset", EventConfigMsg::ResetToDefault),
        ],
    })]
```
This was not possible before since `aligned_column!` did not support using `_` as a placeholder.

Secondary motivation is that this is both a neater API (fewer pub items) and more powerful (it's possible to combine usage of `row!` / `column!` with cell and spanned-cell specifications).

---

Also adds `Layout::Dummy` to reduce secondary errors resulting from a layout parse failure (requires feature `nightly-diagnostics`).

Aside: `nightly-diagnostics` results in warnings on the `Tile::core`, `core_mut` methods. Ideally I would suppress the lint with `#[allow(..)]` here; unfortunately custom proc-macro lints are not yet available.

### DIFF
--- a/crates/kas-macros/src/collection.rs
+++ b/crates/kas-macros/src/collection.rs
@@ -66,18 +66,6 @@ impl Parse for CellInfo {
                 } else {
                     Err(Error::new(lit.span(), format!("expected value >= {start}")))
                 }
-            } else if input.parse::<Token![..]>().is_ok() {
-                let plus = input.parse::<Token![+]>();
-                let lit = input.parse::<LitInt>()?;
-                let n: u32 = lit.base10_parse()?;
-
-                if plus.is_ok() {
-                    Ok(start + n - 1)
-                } else if n > start {
-                    Ok(n - 1)
-                } else {
-                    Err(Error::new(lit.span(), format!("expected value > {start}")))
-                }
             } else {
                 Ok(start)
             }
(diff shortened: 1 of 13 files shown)


dhardy:
Possible further changes to `grid!`:

- Support using `$row` and `$column` to refer to the row/column of the last specified cell and using basic arithmetic (e.g. `$row + 1`).
- Formalise the trick used in `MandlebrotUI` to avoid a margin around the main content of a grid, e.g. by reserving index 0 for margin usage and adding a specifier like `$end`.

Both would be easy enough to implement but increase complexity of the public interface for little gain.

PR #632, opened 2026-01-06
Add plus and minus `MarkStyle`s

Kwarrtz:
This is a naive attempt to add '+' and '-' marks, but the resulting line weight (at least on my system) is inconsistent with the existing mark styles with diagonal strokes, I believe for aliasing reasons
image
Is there likely to be a straightforward solution to this? And if not, is the inconsistency bad enough to be deal-breaking?

Maybe this is unnecessary altogether, since one can just use a `Button` with `"+"` or `"-"` label text, but the default appearance and layout behavior of `Button` and `MarkButton` _is_ different.

### DIFF
--- a/crates/kas-core/src/theme/dimensions.rs
+++ b/crates/kas-core/src/theme/dimensions.rs
@@ -258,7 +258,7 @@ impl<D: 'static> ThemeSize for Window<D> {
                     MarkStyle::Chevron(dir) => match dir.is_vertical() == axis_is_vertical {
                         true => self.dims.mark / 2 + i32::conv_ceil(self.dims.mark_line),
                         false => self.dims.mark + i32::conv_ceil(self.dims.mark_line),
-                    }
+                    },
                     MarkStyle::X | MarkStyle::Plus | MarkStyle::Minus => {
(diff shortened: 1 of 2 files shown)


dhardy:
Using text glyphs all round would be an option, but it would be hard to make consistent, perhaps requiring shipping a specific font with Kas. That said, text glyphs are fast to draw (due to caching) and already have code for sub-pixel rendering and dealing with weight in aliased lines, so might ultimately be the best choice. However getting consistent sizing and alignment may be hard.

A simple fix may be to adjust the line width. `dims.mark_line` is rounded to the nearest integer which makes sense for horizontal and vertical lines; perhaps add `dims.mark_diagonal_line = mark_line / 2.0.sqrt()`.

Kwarrtz:
image

That seems to overcorrect the problem and make the diagonal lines thinner, which is consistent with what I noticed when I had previously tried the opposite (multiplying the width of the horizontal and vertical lines by sqrt(2)). And trying to fine-tune that factor doesn't work, the results are (to my eye at least) identical if I replace sqrt(2) with e.g. 1.1.

I do think this looks better overall though.

dhardy (approved):
LGTM, thanks (but fix the tests)

PR #635, opened 2026-01-12
Enable dual-source blending (requires unreleased wgpu changes)

andyleiserson:


dhardy (approved):
Thanks

Blocker: wgpu release

PR #637, opened 2026-01-15
Remove feature dark-light

dhardy:
Remove the [dark-light](https://crates.io/crates/dark-light) dependency and feature.

Motivation: the crate is not being well maintained, and is causing build failures (in one of its dependencies, `ashpd`, which is not up-to-date).

Further motivation: this small feature required a huge number of additional dependencies. This PR removes the size of the full dependency tree (`cargo tree --all-features|wc -l`) from 1173 to 1048.

Note: further work would be required for run-time switching detection.
Further note: winit already has the `WindowEvent::ThemeChanged` event for run-time detection, though it supports few platforms.


dhardy:
Also: update rfd. We don't need to use a patched version any more. This removes the other `ashpd` dependency.

Also: update usvg, resvg.

PR #642, opened 2026-01-24
IME fix and input/selection focus deconfliction

dhardy:
Relax debug-mode validation of text effects: require increasing but not strictly-increasing order by start index.

Deconflict text selection and input focus: do not request input focus until mouse button release.


dhardy:
A bug not fixed here is that when a text input requests key/selection focus and IME focus simultaneously (without already having selection focus) the IME focus fails since key/selection requests are not handled immediately while the IME request is, and does not find an existing selection focus.

PR #648, opened 2026-01-27
Move most pub &mut self fns out of EventState; apply sel/key focus changes immediately

dhardy:
This moves a number of public `&mut self` fns from `EventState` to either `ConfigCx` or `EventCx`, as is most appropriate for usage.

Also, this PR handles `request_sel_focus` and `request_key_focus` immediately instead of as a pending change. This improves compatibility with `request_ime_focus`, now renamed to `replace_ime_focus` to emphasize that it shouldn't be called repeatedly except to change parameters.


dhardy:
This fixes the bug noted in #642: ime focus can now take effect when an editor without existing selection focus receives focus via touch input.

PR #697, opened 2026-05-31
Editor cleanups / extensions, esp `Rc<String>` support

dhardy:
Some fixes / simplifications for recent PRs (see commits).

The editor types `Common` and `Part` are now pub. New pub fns:

- `Common::direction, set_direction, update_direction, set_colors`
- `Part::status, prepare_wrap`
- `impl<S: ToString> From<S> for Part`
- `Part::text` and `Editor::text` -> `&Rc<String>`
- `Part::clone_text` and `Editor::clone_text` -> `Rc<String>`
- `Editor::set_text(&mut self, cx: _, text: Rc<String>`

### DIFF
--- a/examples/gallery.rs
+++ b/examples/gallery.rs
@@ -21,6 +21,7 @@ use kas::widgets::edit::{EditGuard, Editor, highlight::SyntectHighlighter};
 use kas::widgets::{column, *};
 use kas::window::Popup;
 use std::ops::Range;
+use std::rc::Rc;
 
 #[derive(Debug, Default)]
 struct AppData {
@@ -63,10 +64,10 @@ fn widgets() -> Page<AppData> {
         Check(bool),
         Combo(Entry),
         Radio(u32),
-        Edit(String),
+        Edit(Rc<String>),
         Slider(i32),
         SpinBox(i32),
-        Text(String),
+        Text(Rc<String>),
     }
 
     impl_scope! {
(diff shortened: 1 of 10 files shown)


examples/gallery.rs
```diff
@@ -91,7 +92,7 @@ fn widgets() -> Page<AppData> {
         type Data = Data;
 
         fn activate(&mut self, edit: &mut Editor, cx: &mut EventCx, _: &Data) -> IsUsed {
-            cx.push(Item::Edit(edit.clone_string()));
+            cx.push(Item::Edit(edit.clone_text()));
             Used
```
dhardy, line 96: The gallery can now clone by ref-count instead of copying `String` contents, passing `Rc<String>` as a message

examples/gallery.rs
```diff
@@ -156,7 +155,7 @@ fn widgets() -> Page<AppData> {
         ],
         row![
             "Button",
-            Button::label_msg("&Press me", Item::Button).map_any()
+            Button::label("&Press me").with(|cx, _| cx.push(Item::Button)).map_any()
```
dhardy, line 158: One workaround was required: `Button::label_msg` requires the message type support `Send` (no longer the case due to usage of `Rc`). We can avoid this requirement by direct implementation of the closure.

dhardy (commented):
There is a weak incentive here to use `Arc<String>` instead of `Rc<String>` (since it supports `Send`). I'm not convinced it is sufficient motivation.

A stronger motivation might be to spawn a worker thread doing some work over cloned text contents so this might be revised later. (Note: `&String` *is* `Send` so it may still be possible to send text contents to a worker thread without a deep clone using scoped threads.)

PR #698, opened 2026-06-01
Update wgpu to v0.29, update guillotiere, toml, resvg*

dhardy:
This unblocks #635.


crates/kas-wgpu/src/surface.rs
```diff
@@ -113,22 +113,20 @@ impl<C: CustomPipe> WindowSurface for Surface<C> {
...
+                frame
+            }
+            CurrentSurfaceTexture::Timeout
+            | CurrentSurfaceTexture::Occluded
+            | CurrentSurfaceTexture::Outdated
+            | CurrentSurfaceTexture::Lost
+            | CurrentSurfaceTexture::Validation => {
                 return Instant::now();
             }
         };
```
dhardy, line 128: Inelegant but broadly equivalent to what we had before.

PR #712, opened 2026-09-08
Multi-part text editing

dhardy:
This follows #710, implementing multi-part editing support. There are some pre-existing issues (e.g. mouse interaction with scrolling), but editing does appear to work (basic testing).

Motivation: make the editor scale to large input documents. Performance is acceptable now but could still be better (e.g. I can paste the 2374-line `editor.rs` file into `examples/text-editor` and type away; under a release build there is no noticeable lag while under a debug build some actions like the paste still have noticeable lag though the typing at least is lag-free).

### DIFF
--- a/crates/kas-widgets/src/edit/editor.rs
+++ b/crates/kas-widgets/src/edit/editor.rs
@@ -42,9 +42,9 @@ pub struct ActionResetStatus;
 /// Result type of [`Component::handle_event`]
 #[derive(Debug)]
 pub enum EventAction {
-    /// Key not used, no action
+    /// Input event was not used
     Unused,
-    /// Key used, no action
+    /// Input event was consumed without action
     Used,
     /// Focus has been gained
     FocusGained,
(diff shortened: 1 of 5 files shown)


crates/kas-widgets/src/edit/multi_part.rs
```diff
@@ -251,16 +256,21 @@ mod MultiPartEditor {
...
+            let mut iter = self.text_parts();
+            if let Some(first) = iter.next() {
+                s.push_str(first.as_str());
+            }
+            for part in iter {
+                s.push('\n'); // TODO
+                s.push_str(part.as_str());
             }
             s
         }
```
dhardy, line 276: This doc is wrong; line separators must be inserted. Currently the source line separators are not stored; this code assumes LINE_FEED (`\n`).

I will leave the fix out of this PR since a tweak to `kas_text` wanted (expose the `LineBreakEncoding` type).

crates/kas-widgets/src/edit/editor.rs
```diff
@@ -323,11 +342,66 @@ impl PartList for Part {
...
+    fn remove(&mut self, index: u32) -> Part {
+        Vec::remove(self, index.cast())
+    }
+
+    #[inline]
+    fn delete(&mut self, range: Range<u32>) {
+        let range: Range<usize> = range.cast();
+        Vec::drain(self, range);
+    }
+}
```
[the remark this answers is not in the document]
dhardy: Incorrect: neither the method `Vec::drain` nor `Drain` is marked with `#[must_use]`. Further, the above change does nothing.

Leaking the return-value is problematic but this is not the case here.

dhardy:
> The current multi-part layout/selection paths have confirmed correctness issues (stacking parts using pre-wrap `content_size()` and missing redraw requests after cursor/selection updates).

Addressed, though these issues were not observed in testing.

crates/kas-widgets/src/edit/multi_part.rs
dhardy: Fixed in #713.

PR #715, opened 2026-09-10
Add trait `AutoEditGuard`

dhardy:
# Motivation

Reduce confusion about whether an editors contents is derived from input data or is autonomous.

Note: simple widgets like `CheckBox` require a `state_fn: impl Fn(&ConfigCx, &A) -> bool` or similar so that their state is always driven by input data. For complex widgets like editors this is not always the case. While there is an argument for trying to force this, Kas has a long tradition of supporting stateful widgets; forcing all widgets with "real data" (i.e. not a cache or something like the cursor position) to have an external binding for that data would neither be possible to enforce nor easy to do in a performant fashion.

# Additions

`trait AutoEditGuard` is essentially a cut-down version of `EditGuard` without input data or an `update()` handler.

# Removals

Edit handlers for `configure` and `focus_gained` were removed: `configure` should not normally be wanted (and can be done through an external wrapper like `AdaptEvents` if absolutely required) and `focus_gained` should not (in my opinion) ever be able to influence the state of an editor.

# Changes

`DefaultGuard` is no longer parametrised over the input data type.

Fns `EditBox::with_text` and `EditBoxCore::with_text` are now only available when the guard implements `AutoEditGuard`.

### DIFF
--- a/crates/kas-view/src/filter.rs
+++ b/crates/kas-view/src/filter.rs
@@ -6,7 +6,7 @@
 //! Filters over data
 
 use kas::event::EventCx;
-use kas_widgets::edit::{EditGuard, Editor};
+use kas_widgets::edit::{AutoEditGuard, Editor};
 use std::fmt::Debug;
 
 /// Ability to set filter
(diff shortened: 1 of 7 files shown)


dhardy:
Missing: the provided `StringGuard`, `ParseGuard` and `InstantParseGuard` are all non-auto guards. It may be worth adding auto equivalents. Also, maybe `StringGuard` is redundant.

Edit: see #717.

